PR #14, opened 2015-02-21
Some option parsing fixups when running --search_only

wwood:
This is not a very straightforward merge, I'm afraid. There are two main parts to what I did:
1. take the 0.4.0 version that was on luca not version controlled and use those changes
2. make it so you can just specify a hmm without a refpkg - before the option parsing wasn't working quite right


geronimp:
fixed in next commit

PR #18, opened 2015-02-22
Euk hmm quickfix

wwood:
This isn't a real fix, just needed it for Eleonore to be able to move ahead.


geronimp:
Fixed in next commit

PR #21, opened 2015-03-31
add setup.py so it pip installs, minor code cleanups

wwood:
You can now pip install, but I can't get it to actually run due to problems in the code itself. I imagine those problems are fixed in your branch Joel


wwood:
Note that only python dependencies are installed, a some external dependencies e.g. orfm need to be installed manually.

PR #27, opened 2015-04-29
add untested ability to specify trees in graftm create

wwood:
Hey,

Somehow all the lines seem to have gone to windows line endings again, too.

Actually dunno if this quite works yet, though it is a simple enough change.


geronimp:
From now on I’ll run a bash script that automates the commit to github. This will include a conversion of windows line headings.

PR #44, opened 2015-05-25
Failing test

wwood:
Seems this happens rarely, but sometimes in real data a single fastq sequence can match a HMM in two different frames, seems to cause issues. Do you mind?

ta


wwood:
Oh actually this is different than I thought - the problem is that the HMM is twice hitting the same ORF with a nucleotide input. I'll see if I can fix.

PR #71, opened 2015-07-01
only guess sequence type with the first 300 chars, and require >10% amino acids to call it a protein

wwood:
Came across a genome file where the entire genome was on the second line of the file, and there was 2 non-ATGCNU characters in the 4 Mbp or so, so it was calling it a protein sequence.


geronimp:
Thanks Ben

PR #73, opened 2015-07-05
Geronimp diamond

geronimp:


geronimp:
This will be v.0.7.0 

Let me know what you think of the changes and I'll merge.
Joel

wwood:
not a fast forward, can you do the final merge please?

geronimp:
Yep, just needed your green light.


Joel


wwood:
No I mean merge to get to another pull request.

wwood:
Still doesn't appear to be automatically mergable on github, looks like your master branch was out of date when you merged it

wwood:
Sorry, let me have a look at this before you merge.

bin/graftM
```diff
@@ -78,6 +78,10 @@ UTILITIES
 =====================================================================================================
 """ % (graftm.__version__)
 
+def header():
```
wwood: A better name for this method would be `print_header` but eh

graftm/hmmer.py
```diff
@@ -26,57 +42,54 @@ def __init__(self, search_hmm, aln_hmm=None):
...
-            Output alignment of forward reads with insertions removed
-        rev_conv_file : str
-            Output alignment of reverse reads with insertions removed
-        
+        directions : dict
+            dictionary containing readnames as keys, and compliment direction
+            as the entry (False=Forward, True=Reverse)
         Returns
         -------
         Nothing - output files are known.
```
wwood: this is now untrue, please amend.

graftm/hmmer.py
```diff
@@ -26,57 +42,54 @@ def __init__(self, search_hmm, aln_hmm=None):
...
-            Output forward compliment path.
-        rev_file : str
-            Output reverse compliment path.
-        for_conv_file : str
-            Output alignment of forward reads with insertions removed
-        rev_conv_file : str
-            Output alignment of reverse reads with insertions removed
-        
+        directions : dict
+            dictionary containing readnames as keys, and compliment direction
```
wwood: complement I think you meant. But it is anti-intuitive to have fwd=False. Is it hard to switch?

graftm/hmmer.py
```diff
@@ -170,10 +185,10 @@ def hmmsearch(self, output_path, input_path, unpack, seq_type, threads, eval, or
             raise Exception('Programming Error: error guessing input sequence type')
         
         # Run the HMMsearches
-        searcher = HmmSearcher(threads, eval)
+        searcher = HmmSearcher(threads, '-E %s' % eval)
```
wwood: please change the name eval to evalue

graftm/hmmer.py
```diff
@@ -318,10 +353,10 @@ def __check_euk_contamination(self, hmm_hit_tables, run_stats):
         
         Returns
         -------
-        run_stats : dict
-            Updated form of the above.
+        euk_reads : array
```
wwood: this line is unecessary since the name of the variable that is returned has no meaning outside this function

graftm/hmmer.py
```diff
@@ -414,7 +422,7 @@ def removeOverlaps(item):
                     continue
             return item
                     
-        def extractMultipleHits(reads_path, stats):
+        def extractMultipleHits(reads_path, stats): # TODO: Not used for the time being - need a more robust method
```
wwood: Not sure I quite follow your TODO comment. Are you saying it no longer works?

graftm/hmmer.py
```diff
@@ -507,56 +500,57 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
                         output_file.write(fasta_id)
                         output_file.write(fasta_seq)
 
-    def extract_orfs(self, input_path, raw_orf_path, hmmsearch_out_path, orf_titles_path, orfm, orf_out_path):
+    def extract_orfs(self, input_path, orfm, orf_out_path):
         '''
-        orfm: graftm.OrfM object with parameters already set
+        extract_orfs - Extract only the orfs that hit the hmm, return sequence 
+                       file with within.
```
wwood: How about "Return a path to a FASTA file containing ORFs that hit the HMM (self.search_hmm)

graftm/hmmer.py
```diff
@@ -507,56 +500,57 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
...
-            for table in output_table_list:
-                for title in [x.split(' ')[0] for x in open(table).readlines() if not x.startswith('#')]:
-                    if title not in seen:
-                        output.write(str(title) + '\n')
-                        seen.append(title)
-                    else:
-                        pass       
-        
+            reads= set(
+                       sum([HMMreader(x).names() for x in output_table_list],
```
wwood: This would be better written with itertools.chain

graftm/hmmer.py
```diff
@@ -507,56 +500,57 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
...
-        # Return name of output file
-        return orf_out_path
-
-    def p_search(self, files, args, run_stats, base, unpack, raw_reads):
-        '''Protein search pipeline - The searching step for the protein 
+    
+    @timeit
+    def p_search(self, files, args, base, unpack, raw_reads, search_method, gpkg):
+        '''
+        Protein search pipeline - The searching step for the protein
```
wwood: "Protein" is confusing here, I never remember whether you mean searching protein sequences or searching with a protein HMM/diamonddb

graftm/hmmer.py
```diff
@@ -568,28 +562,28 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
         base : str
             The name of the input file, stripped of all suffixes, and paths. 
             Used for creating file names with 'files' object.
+        unpack : obj
+            unpack_sequences.py object, returns string command that will output
```
wwood: Best to specify the class name, not the file name. Also typo in sequences

graftm/hmmer.py
```diff
@@ -568,28 +562,28 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
+            sequenecs to stdout when called on command line (use: unpack.command_line() )
         raw_reads : str
             The reads to be searched.
-
+        search_method : str
+            The method for searching e.g. 'hmmsearch' or 'diamond'
```
wwood: maybe either not e.g.

graftm/hmmer.py
```diff
@@ -598,63 +592,87 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
+                                           )
+            for result in search_result:
+                hit_readnames=[x[0] for x in result.each([SequenceSearchResult.QUERY_ID_FIELD])]
+        
+        elif search_method == 'diamond':
+            #run diamond
+            search_result =  Diamond(
+                                     database=os.path.join(
+                                                           gpkg._base_directory, \
+                                                           gpkg._contents_hash[gpkg.DIAMOND_DATABASE_KEY]
```
wwood: Don't use single underscore methods from outside the class they are defined in. Here there is a method for getting the diamond database

graftm/hmmer.py
```diff
@@ -598,63 +592,87 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
+            self._extract_from_raw_reads(
+                                        hit_reads_fasta,
+                                        readnames.name,
+                                        raw_reads,
+                                        unpack.format()
+                                        )
 
         if not hit_readnames:
-            return False, run_stats
+            return None, None, None
```
wwood: This happens when there are no search_hmms? If so I would just raise an error rather than trying to carry on

graftm/hmmer.py
```diff
@@ -598,63 +592,87 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
+            self.extract_orfs(hit_reads_fasta,
+                              extracting_orfm,
+                              hit_reads_orfs_fasta)
+            
+            hit_reads_fasta=hit_reads_orfs_fasta
+        
+        ## TODO: This method does not work when searching a genome, or assembled
+        ## data, because there may be > 1 copy of gene, and very spaced. Simply
+        ## choosing one would not suffice in that context. Will have to rewrite
+        ## So that multiple hits with same read name are delt with properly.
```
wwood: Unfortunately I feel this needs to be fixed before we can merge this branch

graftm/hmmer.py
```diff
@@ -598,63 +592,87 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
+                                        readnames.name,
+                                        raw_reads,
+                                        unpack.format()
+                                        )
 
         if not hit_readnames:
-            return False, run_stats
+            return None, None, None
         
         if args.input_sequence_type == 'nucleotide':
```
wwood: best to read this data from unpack, it would be best to try to move away from using args in this class

graftm/hmmer.py
```diff
@@ -675,88 +693,83 @@ def d_search(self, files, args, run_stats, base, unpack, raw_reads, euk_check):
             The reads to be searched.
         euk_check : bool
             True False, whether to check the entire sample for euk reads.
```
wwood: "True indicates the sample will be checked for eukaryotic reads, False indicates not."

graftm/hmmer.py
```diff
@@ -598,63 +592,87 @@ def p_search(self, files, args, run_stats, base, unpack, raw_reads):
...
-        run_stats['euk_contamination'] = 'N/A'
-        run_stats['euk_uniq'] = 'N/A'
-        run_stats['euk_check_t'] = 'N/A'
-        # Return hit reads, and summary hash
-        return hit_orfs, run_stats
-
-    def d_search(self, files, args, run_stats, base, unpack, raw_reads, euk_check):
-        '''Nucleotide search pipeline - The searching step for the nucleotide
+        hit_read_counts=[0,len(hit_readnames)]
+        return hit_reads_fasta, search_result, hit_read_counts
```
wwood: It seems a bit clunky to return several things, where one of those is several things. Perhaps return an instance of a new class that includes these info?

graftm/hmmer.py
```diff
@@ -675,88 +693,83 @@ def d_search(self, files, args, run_stats, base, unpack, raw_reads, euk_check):
...
+                                                    args.eval
+                                                    )
+            for result in search_result:
+                hit_readnames=[x[0] for x in result.each([SequenceSearchResult.QUERY_ID_FIELD])]
             
-        Raises
-        ------
-        N/A
+        elif search_method == 'diamond':
+            logging.error("Diamond searches not supported for nucelotide sequences yet")
```
wwood: typo. Also why log an error if you are raising an exception anyway? So just delete this line. Also unclear whether you mean nucleotide query or database

graftm/hmmer.py
```diff
@@ -675,88 +693,83 @@ def d_search(self, files, args, run_stats, base, unpack, raw_reads, euk_check):
...
         
-        Examples
-        --------
-        N/A
+        Parameters
+        ----------
+        input_path : str
+        output_path : str
+        reverse_direction : dict
+            A dictionary of read names, with the entries being the compliment
```
wwood: complement

graftm/housekeeping.py
```diff
@@ -129,18 +122,15 @@ def parameter_checks(self, args):
...
             self._check_file_existence(args.forward)
-            if hasattr(args, 'reverse'):
+            if args.reverse:
                 self._check_file_existence(args.reverse)
                     
             # Determine the File format based on the suffix
             
             sequence_file_list = []
-            if hasattr(args, 'reverse'):
+            if args.reverse:
```
wwood: excellent, thanks for doing this.

graftm/pplacer.py
```diff
@@ -3,12 +3,22 @@
...
 import timeit
 import logging
+import time
 
 from Bio import SeqIO
 
 from graftm.classify import Classify
 from graftm.housekeeping import HouseKeeping
 
+def timeit(method):
```
wwood: this method is used in both hmmer and pplacer (at least). Don't repeat yourself. Make a new class that has this method and use that class from both places, perhaps make it a staticmethod for ease of use.

wwood: I don't know, does using this slow down the code significantly? An alternative might just be to print out timestamps

graftm/pplacer.py
```diff
@@ -138,17 +187,16 @@ def check_reads(read_lists):
...
             return False
         # Read in the read names for each file
-        check = check_reads([hash['forward']['reads'].keys(), hash['reverse']['reads'].keys()])
+        check = check_reads([forward_reads, reverse_reads])
         if check:
-            forward_read_names = [x.replace('/1', '') for x in hash['forward']['reads'].keys()]
-            reverse_read_names = [x.replace('/2', '') for x in hash['reverse']['reads'].keys()]
+            forward_read_names = set([x.replace('/1', '') for x in forward_reads])
+            reverse_read_names = set([x.replace('/2', '') for x in reverse_reads])
         elif not check:
```
wwood: isn't this just 'else'?

graftm/pplacer.py
```diff
@@ -160,15 +208,15 @@ def check_reads(read_lists):
...
+    def compare_placements(self, forward_gup, reverse_gup, crossover, placement_cutoff):
         ## Take guppy placement file for the forward and reverse reads, compare
         ## the placement, and make a call as to which is to be trusted. Return
         ## a list of trusted reads for use by the summary step in GraftM
         # Report and record trusted placements
         comparison_hash = {'trusted_placements': {}} # Set up a hash that will record info on each placement
-        for read in hash['crossover']:
+        for read in crossover:
             if read.startswith('FCC'):
                 f_read = read + '/1'
```
wwood: I know it isn't from this merge, but this seems dangerous, I'm not sure all /1 style names start with FCC. Maybe a better way to do this would be to record in the reads are of the '/1' type when you have the full read name, and then refer to that info here.

graftm/run.py
```diff
@@ -37,149 +40,72 @@ def setattributes(self, args):
             self.sequence_pair_list = self.hk.parameter_checks(args)
             if hasattr(args, 'reference_package'):
                 self.p = Pplacer(self.args.reference_package)
+    
+    def _get_sequence_directions(self, search_result):
```
wwood: this method really tested my python understanding (particularly since there are no comments). Are you trying to make a dictionary that maps query IDs to alignment directions? If so isn't a simple for loop better, and allows you to more easily deal with name collisions?

graftm/run.py
```diff
@@ -37,149 +40,72 @@ def setattributes(self, args):
...
-        hit_reads, run_stats = self.h.d_search(self.gmf,
-                                               self.args,
-                                               run_stats,
-                                               base,
-                                               unpack,
-                                               sequence_file,
-                                               summary_dict['euks_checked'])
+    def summarise(self, base_list, trusted_placements, reverse_pipe, times, hit_read_count_list):
+        '''
+        summarise -
```
wwood: please add a comment

graftm/run.py
```diff
@@ -37,149 +40,72 @@ def setattributes(self, args):
...
+        '''
+        summarise - 
         
-        if not hit_reads:
-            return summary_dict, False
+        Parameters
+        ----------
+        base_list : list
+        trusted_placements : dict
+        reverse_pipe : bool
```
wwood: this list incomplete, I was particularly intrigued to know what 'times' meant

wwood:
Hey,

So I think I've finished commenting for the moment, but I haven't run any of the code yet, to test it out. Unfortunately we need to solve the multiple hits issue before this can be merged, I think - too much of a regression. It would break singlem for instance.

geronimp:
Hey, so to fix these issues, do I just keep working on my diamond branch and merge again?


Joel

wwood:
Yeh, although do a pull request and don't just merge please.

graftm/hmmer.py
geronimp: fwd=False to follow the logic in the check on line 68, which is checking if any reads are in the reverse compliment (if any(True) = True) If it were the other way around, (i.e. if any(False) = False), I would need to invert the list anyway to get past the check.

wwood: why not
```
any([direction == False for direction in directions.values()])
```
though? Maybe this is too hard though as you'd want to change the HMM table parser, which I think you use outside graftM now.

wwood: Or better yet to not store the entire thing in memory use round brackets
```
any((direction == False for direction in directions.values()))
```

graftm/hmmer.py
geronimp: Not sure I follow.  euk reads are used in the d_search pipeline

graftm/hmmer.py
geronimp: Yes this was a regression that I forgot to fix up. I'll see if I can fix it before I submit another pull request

graftm/hmmer.py
wwood: I just mean the line in the documentation. The fact that the variable returned is named `euk_reads` is not relevant outside this function. I would just say
```
Returns
-------
array of all reads names deemed to be eukaryotic
```

graftm/hmmer.py
geronimp: GraftM will fail if you don't have search HMMs I'm fairly certain. Here we continue if no hits are found because we're in a for loop iterating through the files given to graftM (the sequences being searched). So if one happens to have none, why should the whole graftM run fail?

graftm/hmmer.py
geronimp: Agreed - will do

graftm/hmmer.py
geronimp: oh, gotcha

graftm/hmmer.py
wwood: Oh ok. I think then it might be better to return an empty data lists/etc, as opposed to an actual `None`, because the None complicates downstream code, perhaps. I'm not entirely sure though.

graftm/hmmer.py
geronimp: Or even better yet,
```
if False in direction.values()
```
which is easier to read anyway

It should be easy to make the change.

wwood: tru dat

graftm/pplacer.py
geronimp: doesn't appreciably slow code, no. Just fixed.

PR #78, opened 2015-07-16
Geronimp pre files

geronimp:


geronimp:
doesn't print header now

bin/graftM
```diff
@@ -117,6 +117,8 @@ if __name__ == '__main__':
     placement_options.add_argument('--placements_cutoff', metavar='confidence', help='Cutoff of placement confidence level (0.5 - 1), default = 0.75', default=0.75)
     placement_options.add_argument('--resolve_placements', action="store_true", help='Force taxonomy down to the best placement', default=False)
     placement_options.add_argument('--merge_reads', action="store_true", help='Merge forward and reverse read alignments for placement')
+    placement_options.add_argument('--cluster', action="store_true", help='Cluster reads at 100%% before placement to reduce tree insertion time WARNING: only use with unassembled data (i.e. equel read length)')
```
wwood: typo equel

graftm/db_search_results.py
```diff
@@ -0,0 +1,48 @@
+from graftm.deduplicator import Deduplicator
+from graftm.sequence_io import SequenceIO
+
+import logging
+import os
+
+class DBSearchResult:
+    '''Class for containing results from search pipeline in GraftM '''
+    def __init__(self, ouput_reads, search_result, hit_read_count, slash_endings):
```
wwood: some more comments for this class would be good I think, given it is a new class

graftm/graftm_package.py
```diff
@@ -63,7 +70,7 @@ def acquire(graftm_package_path):
         return pkg
     
     def check_universal_keys(self, version):
-        h = self._contents_hash
+        h = self.contents_hash
```
wwood: why was the underscore removed here and for base_directory?

graftm/hmmer.py
```diff
@@ -52,20 +45,23 @@ def hmmalign(self, input_path, directions):
         input_path : str
             Filename of unaligned hits to be aligned
         directions : dict
-            dictionary containing readnames as keys, and compliment direction
-            as the entry (False=Forward, True=Reverse)
+            dictionary containing read names as keys, and compliment
```
wwood: compliment typo

wwood: and isn't this the direction of the read, not 'complement'?

graftm/hmmer.py
```diff
@@ -52,20 +45,23 @@ def hmmalign(self, input_path, directions):
...
+            as the entry (True=Forward compliment, False=Reverse compliment)
         Returns
         -------
-        Nothing - output files are known.
+        conv_files : array
```
wwood: No need to say `conv_files`. And while I'm here, to be a bit more specific maybe
```
list of two strings, where each string is a path to an aligned fasta file (forward and reverse files respectively)
```

graftm/hmmer.py
```diff
@@ -93,9 +89,9 @@ def hmmalign(self, input_path, directions):
             with open(for_file, 'w') as for_aln:
                 logging.debug("Writing forward compliment reads to %s" % for_file)
                 for record in forward:
-                    if record.id and record.seq: # Check that both the sequence and ID fields are there, HMMalign will segfault if not.
-                        for_aln.write('>'+record.id+'\n')
-                        for_aln.write(str(record.seq)+'\n')
+                    if record.id and record.seq:  # Check that both the sequence and ID fields are there, HMMalign will segfault if not.
```
wwood: I'm a little confused by this. Why would these not be there? And is it a good idea to ignore them if they aren't?

graftm/hmmer.py
```diff
@@ -142,16 +141,16 @@ def hmmsearch(self, output_path, input_path, unpack, seq_type, threads, eval, or
         input_path : str
             path to input sequences to search
         unpack : obj
-            Object that builds the commad chunk for unpacking the raw sequences.
-            First guesses file format, and unpacks appropriately. Calls 
-            command_line to construct final command line string.
+            UnpackRawReads object, returns string command that will output
```
wwood: instead of `obj`, just say `UnpackRawReads`

graftm/hmmer.py
```diff
@@ -405,51 +367,26 @@ def _extract_from_raw_reads(self, output_path, input_path, raw_sequences_path, i
...
-                    out_reads[key]=str(reads[key].seq)
-                    new_stats[key]=item[0]
-            out_path = reads_path[:-3]+'_split.fa'
-            with open(out_path, 'w') as out:
-                for key,item in out_reads.iteritems():
-                    out.write(">%s\n" % (str(key)))
-                    out.write("%s\n" % (str(item)))
-            return new_stats, out_path
-        
+        def extractMultipleHits(stats, reads_path):
```
wwood: as this is a new method can you please document.

graftm/hmmer.py
```diff
@@ -462,10 +399,16 @@ def extractMultipleHits(reads_path, stats): # TODO: Not used for the time being
             cmd = "%s %s | awk '{print \">\" substr($0,2);getline;print;getline;getline}' > %s" % (fxtract_cmd, raw_sequences_path, output_path)
         else:
             raise Exception("Programming error")
-        
         logging.debug("Running command: %s" % cmd)
         subprocess.check_call(cmd, shell=True)
-
+        
+        # If any of the hits have more than one alignment span (i.e. >1 hit)
+        if any([x for x in hits.values() if len(x) > 1]):
```
wwood: I think round brackets would be better than square brackets here - square brackets means the whole thing is in memory

wwood: In fact what is the point of this if statement? Why not just always extractMultipleHits?

graftm/hmmer.py
```diff
@@ -545,144 +490,212 @@ def extract_orfs(self, input_path, orfm, orf_out_path):
...
         subprocess.check_call(cmd, shell=True)
     
-    @timeit
-    def p_search(self, files, args, base, unpack, raw_reads, search_method, gpkg):
+    def _get_read_names(self, search_result, max_range):
         '''
-        Protein search pipeline - The searching step for the protein 
-        pipeline, where hits are identified in the input reads through 
-        searches with hmmer or diamond.
+        _get_read_names - loops through hmm hits and their alingment spans to
```
wwood: I think there is no need to repeat the name of the method - it is right there already.

graftm/hmmer.py
```diff
@@ -545,144 +490,212 @@ def extract_orfs(self, input_path, orfm, orf_out_path):
...
         elif search_method == 'diamond':
-            #run diamond
-            search_result =  Diamond(
+            # run diamond
+            search_result = Diamond(
                                      database=os.path.join(
-                                                           gpkg._base_directory, \
-                                                           gpkg._contents_hash[gpkg.DIAMOND_DATABASE_KEY]
+                                                           gpkg.base_directory, \
+                                                           gpkg.contents_hash[gpkg.DIAMOND_DATABASE_KEY]
```
wwood: gpkg should not be used in this way, use `diamond_database_path()`

graftm/hmmer.py
```diff
@@ -545,144 +490,212 @@ def extract_orfs(self, input_path, orfm, orf_out_path):
...
             raise Exception("Programming error: unexpected search_method %s" % search_method)
+        
         with tempfile.NamedTemporaryFile(prefix='graftm_readnames') as readnames:
-            orfm_regex = re.compile('^(\S+)_(\d+)_(\d)_(\d+)')
-
+            # Write the names of hits to a tmpfile
+            
+            orfm_regex = re.compile('^(\S+)_(\d+)_(\d)_(\d+)')  # to remove OrfM suffix from read names
+            if srch_aln_only:
+                hits = {key:[] for key in list(
```
wwood: There's unecessary `list()`s here. But more to the point a regular for loop would be much easier to understand. e.g.
```
for result in search_result:
for array in result.each([SequenceSearchResult.QUERY_ID_FIELD]):
hits[array[0]] = []
```

graftm/hmmer.py
```diff
@@ -545,144 +490,212 @@ def extract_orfs(self, input_path, orfm, orf_out_path):
...
+                                               itertools.chain(
+                                                               *itertools.chain(
+                                                                                *[list(result.each([SequenceSearchResult.QUERY_ID_FIELD])) \
+                                                                                  for result in search_result]
+                                                                                )))}
+            else:
+                hits = self._get_read_names(search_result,  # define the span of hits
+                                            gpkg.contents_hash[gpkg.RANGE_KEY])
```
wwood: as above, a new method is needed in the gpkg object I guess.

graftm/hmmer.py
```diff
@@ -706,47 +723,66 @@ def d_search(self, files, args, base, unpack, raw_reads, euk_check, search_metho
...
+
+                
         elif search_method == 'diamond':
-            logging.error("Diamond searches not supported for nucelotide sequences yet")
-            raise Exception("Diamond searches not supported for nucelotide sequences yet")
-        
+            raise Exception("Diamond searches not supported for nucelotide databases yet")
+
         with tempfile.NamedTemporaryFile(prefix='graftm_readnames') as readnames:
+            if srch_aln_only:
```
wwood: as above

graftm/hmmer.py
```diff
@@ -756,7 +792,7 @@ def align(self, input_path, output_path, directions):
         input_path : str
         output_path : str
         reverse_direction : dict
-            A dictionary of read names, with the entries being the compliment 
+            A dictionary of read names, with the entries being the complement
```
wwood: s/complement/direction/

graftm/unpack_sequences.py
```diff
@@ -8,7 +11,51 @@
...
+        num_nucleotide = 0
+        num_protein = 0
+        count = 0
+        for residue in seq:
+            if residue in nas:
+                num_nucleotide += 1
+            elif residue in aas:
+                num_protein += 1
+            else:
+                raise Exception(logging.error('Encountered unexpected character when attempting to guess sequence type: %s' % (residue)))
```
wwood: not come across this style before, what does logging.error return?

graftm/create.py
```diff
@@ -198,6 +203,44 @@ def cleanup(self, the_trashcan):
             else:
                 os.remove(every_piece_of_junk)
     
+    def _define_range(self, sequences):
+        '''
+        define_range - define the maximum range within which two hits in a db 
+        search can be linked. This is defined as 1.5X the average length of all
```
wwood: I wonder if this definition is sufficient. Maybe 1.5X the length of the maximal sequence length? What do you think?

graftm/create.py
```diff
@@ -198,6 +203,44 @@ def cleanup(self, the_trashcan):
...
+        Returns
+        -------
+        max_range : int 
+            As described above, 1.5X the size of the average length of genes 
+            within the database
+        '''
+        sequence_count = 0
+        total_sequence = 0
+        
+        cmd = 'cat %s' % sequences # output the sequences to STDOUT
```
wwood: I think it would be better to just use a normal fasta format reader here. Writing one from scratch may lead to problems. Is it really that much faster?

graftm/create.py
```diff
@@ -181,14 +182,18 @@ def _compile(self, base, refpkg, hmm, diamond_database_file, prefix):
...
-            "rfpkg": refpkg,
-            "TC":False,
-            "diamond_database": diamond_database_file_in_gpkg}
+                    "aln_hmm": hmm_file_in_gpkg,
+                    "search_hmm": [hmm_file_in_gpkg],
+                    "rfpkg": refpkg,
+                    "TC":False,
+                    "diamond_database": diamond_database_file_in_gpkg,
+                    "range": max_range,
+                    "tax_info": taxonomy_file_in_refpkg}
```
wwood: perhaps I missed it, but as far as I can see this is not used at all - is it needed?

geronimp: Yes, it is needed for the classify module, but somewhere down the track I didn't plug it in.

wwood: ok well I think not abbreviating this is a good idea e.g. calling it `taxtastic_taxonomy_file` or similar

graftm/create.py
geronimp: No, reading in from stdout just makes the calculation below more simple. 

Is that SeqIO module you have good to use?

wwood: Could do. But then there's no reason not to use biopython? Then like
```
for s in SeqIO.read(open(sequences)): 
total_sequence+=1
sequence_count+=len(s.seq)
```

graftm/hmmer.py
geronimp: Not sure, I might have put this in a while ago, so for now I'll remove and see if I run into any segfaults like it says.

geronimp: I agree though not good to ignore

graftm/hmmer.py
geronimp: extractMultipleHits reads in all of the sequences with biopython, which can take time. If all the reads are hit only once, then it is better to skip this process if they are just being re-written to file anyways.

graftm/hmmer.py
geronimp: I'm very hack and slash sometime aren't I?

graftm/unpack_sequences.py
geronimp: Actually just a mistake

geronimp:
running nosetests on changes now. If they pass this should be good to go.

geronimp:
Sorry, didn't mean to close then

geronimp:
Passes nosetests now

wwood:
OK so shall I review then?

geronimp:
lemme clone my repository and and rerun nosetests to confirm, ill let you know in a sec


geronimp:
Try now


wwood:
OK I think we are good with the code. However, I'm a bit leery of pushing this to the master branch / make a release when there is no graftm packages, yeh?

While that is being done let's make changes to the code from this branch.

PR #82, opened 2015-07-26
add bootstrap mode

wwood:
Incidentally, I also removed `default=argparse.SUPPRESS` from a graft option, and fixed a bug that happened when maximum_range was `None`.

If you are happy with these changes then go ahead and merge.
ta.


geronimp:
Thanks for that bug fix

PR #83, opened 2015-07-27
Fxtract piping and a few other bug fixes.

geronimp:


geronimp:
See what you think

geronimp:
whoa, hang on

PR #84, opened 2015-07-27
See howt his goes

geronimp:


graftm/hmmer.py
```diff
@@ -453,52 +453,35 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
...
-    def extract_orfs(self, input_path, orfm, orf_out_path):
+
+    def _extract_orfs(self, input_path, orfm, hit_readnames, output_path):
         '''
-        extract_orfs - Return a path to a FASTA file containing ORFs that hit 
-        the HMM (self.search_hmm)
+        Provide the read names of orfM orfs that hit, the nucleotide sequences
+        and the orfM command line. First writes the readnames to a tmp file,
+        to be used by fxtract to extract sequences. Then run command, which 
+        calls orfs, and pipes them to fxtarct, which extracts the reads of the
```
wwood: typo

graftm/hmmer.py
```diff
@@ -453,52 +453,35 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
...
-        # Extract the reads using the titles.
-        cmd = 'fxtract -H -X -f %s %s > %s' % (orf_titles_path, raw_orf_path, orf_out_path)
-        logging.debug("Running command: %s" % cmd)
-        subprocess.check_call(cmd, shell=True)
+        hit_readnames : str
+            path to a file containin the readnames of hits to the HMM, one per 
+            line.
+        '''        
+        # Write hit readnames to file
+        with tempfile.NamedTemporaryFile(prefix='graftm_readnames') as orfm_readnames:
```
wwood: I think it might have been preferable to use subprocess.pipe or whatever it is to write to the process directly rather than using a tempfile, but eh.

graftm/hmmer.py
```diff
@@ -588,6 +572,31 @@ def _get_read_names(self, search_result, max_range):
...
+        '''
+        Provide a list of readnames to be checked for the /1 or /2 endings which
+        cause troubles downstream when working with paired reads. This function
+        checks the list for ANY read ending in /1 or /2. The reads are first 
+        stripped of any comment lines, using the 'split' function
+        
+        Parameters
+        ----------
+        readnames : list
+            list of strings, each a reanames to be checked for the /1 or /2
```
wwood: typo

graftm/hmmer.py
```diff
@@ -588,6 +572,31 @@ def _get_read_names(self, search_result, max_range):
...
+        Provide a list of readnames to be checked for the /1 or /2 endings which
+        cause troubles downstream when working with paired reads. This function
+        checks the list for ANY read ending in /1 or /2. The reads are first 
+        stripped of any comment lines, using the 'split' function
+        
+        Parameters
+        ----------
+        readnames : list
+            list of strings, each a reanames to be checked for the /1 or /2 
+            ending
```
wwood: what's the return?

wwood:
looks good beyond those minor things.

graftm/hmmer.py
geronimp: Never used that method before, sorry didn't know about it.

geronimp:
mmkay, try now. The subprocess pipe I didn't change, I need an explanation because I dont fully understand sorry

geronimp:
Just added another bugfix too.

graftm/hmmer.py
wwood: something like this
```
In [12]: s = subprocess.Popen(['head','-n2'], stdin=subprocess.PIPE)

In [13]: s.communicate('1\n2\n3\n4\n')
1
2
Out[13]: (None, None)
```
I think you can use subprocess.check_output instead of Popen.

geronimp: Still not sure sorry, I'm not seeing how this fits in with the fxtract command below sorry. I'm writing the readnames to file so they can be passed to the -f flag in fxtract.

wwood: (untested)
```
process = subprocess.check_call('fxtract -H -X -f /dev/stdin <(%s %s) > %s' % (orfm_cmd, input_path, output_path, stdin=PIPE)
process.communicate('\n'.join(hit_readnames))
```
That way there is no need for tempfiles even

wwood:
hey is this ready to merge now? Guess there seems to be conflicts.

geronimp:
I've fixed the conflicts merge when you're ready.
Joel

graftm/hmmer.py
```diff
@@ -449,50 +456,34 @@ def alignment_correcter(self, alignment_file_list, output_file_name):
...
+        ''' 
+        # Write hit readnames to file
+        with tempfile.NamedTemporaryFile(prefix='graftm_readnames') as orfm_readnames:
+            for readname in hit_readnames: 
+                orfm_readnames.write(readname + '\n')
+            orfm_readnames.flush()
+            # Build and run command to extract ORF sequences:
+            orfm_cmd = orfm.command_line()
+            cmd = 'fxtract -H -X -f %s <(%s %s) > %s' % (orfm_readnames.name, orfm_cmd, input_path, output_path)
+            extern.run(cmd)
```
wwood: Hmm, so a few things here still sorry.
* use Popen.communicate, rather than writing a tempfile
* `output_path` is not documented
* it's OrfM, not orfM.
* The first blurb doesn't say what the function actually does, only how it works. Maybe
```
Call ORFs on a nucleotide file and extract the proteins whose name is in `hit_readnames`.
```
How it works isn't so relevant.

graftm/hmmer.py
```diff
@@ -511,10 +502,10 @@ def _get_read_names(self, search_result, max_range):
             alignment span.
         max_range : int
             Maximum range that a gene can extend within a contig. Any hits 
-            that extend beyond this length cannot be linked. max_range is defined 
-            as 1.5 X the average length of all full length genes used in the 
-            search database. This is defined in the CONTENTS.json file within a 
-            gpkg.
+            that extend beyond this length cannot be linked. max_range is 
+            defined as 1.5 X the average length of all full length genes used
```
wwood: `defined` doesn't seem like the right word here. It is either defined, or can be set in the gpkg. Also isn't used in this function that I can see.

wwood:
A few more things to fix I think sorry

graftm/hmmer.py
geronimp: Okay, I changed it to "set" to avoid ambiguity.

It's used below on line 570 and 574

geronimp:
Try now?

wwood:
Thanks.

geronimp:
Are we merging to master now?

wwood:
OK, let's.

wwood:
I think you might have to `git push -f`

geronimp:
Okay, try now


wwood:
merged with forced update to master

PR #86, opened 2015-07-27
standardise the bootstrap HMM output file

wwood:
This is just so the bootstrap HMM is now `bootstrap.hmm` in the base of the output directory.


geronimp:
Cool

PR #89, opened 2015-08-03
remove graftm assemble for conciseness of code

wwood:
not sure about extract_sequences?


geronimp:
Extract sequences can go too. Maybe also the "filter" pipeline which is still hanging around in the graftm script in bin/

wwood:
hey if you like then please merge

PR #94, opened 2015-08-07
be more conservative with defining the basename of samples to avoid clashes

wwood:
with singlem, a lot of people are running lots of recovereged genomes through graftm. Problem is, many of them are named like this
```
binnning1.bin1.fasta
binnning1.bin2.fasta
```
The current basename strategy in graftM collapses these both to `binning1`, and that causes issues. This fix makes them e.g. `binning.bin1`. However, this also means `20120600_E1M.1.fq.gz` -> `20120600_E1M.1` not `20120600_E1M` as before. I think this tradeoff is worth it, you?

I also changed the `--verbosity` flags in test_graft.py to be 2 so when they fail they give more informative errors.

### DIFF
--- a/graftm/graftm_output_paths.py
+++ b/graftm/graftm_output_paths.py
@@ -2,11 +2,11 @@ import os
 
 class GraftMFiles:
     
-    def __init__(self, old_title, outdir, direction):   
+    def __init__(self, old_title, outdir, direction):
         if direction == 'forward' or direction == 'reverse':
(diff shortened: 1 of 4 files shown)


geronimp:
Hmm, I would argue that the naming syntax of the bins below is more of a problem, but maybe its just me that don't like putting '.' in the middle of files so I'll hold my tongue :). Besides since we already sort of use the strategy you have implemented below when we determine the file format I think this change is perfectly fine.


geronimp:
Passes all tests yea?

wwood:
yeh

PR #97, opened 2015-08-21
Geronimp diamond issue

geronimp:
this branch now calls the orfs diamond picked up. They match orfm orfs, but some come out shorter, depending on alignment information provided by diamond.

Check and merge please!


wwood:
Hey, thanks for that.

Couple things. I reckon we want the "alignment" to actually be an alignment, so I reckon we want to hmmalign the "ORFs" that diamond produces.

So following on, there is a method `align` in `hmmer.py` where there is a `pipeline`. I'm thinking there should only be one pipeline (the hmmalign one), and the ORFs called should be fed to this method.

Also the test you added is only --search_and_align_only, I imagine without this flag it will fail since it isn't a real alignment. I wrote some tests for this in my other pull request, maybe you can steal them from there.

geronimp:
I dont follow. Reads are still aligned when --search_and_align_only is not specified. I wrote the test to check the orf calling was okay.



Joel

geronimp:
also, a hmm alignment of the orfs is produced, even if you do pass it --search_and_align_only

wwood:
Oh sorry I misunderstood. When I fix the merge conflicts with my branches I'll get back to you.

graftm/hmmer.py
```diff
@@ -925,7 +960,7 @@ def search_and_extract_nucleotides_matching_nucleotide_database(self,
         return result
         
     @T.timeit
-    def align(self, input_path, output_path, directions):
+    def align(self, input_path, output_path, directions, pipeline):
```
wwood: hey can you add comment for the new parameter?

geronimp: good call

PR #103, opened 2015-08-26
bugfix: when hmmsearch hits but diamond has no hits, assign to Root

wwood:
damn real data.


geronimp:
So if it recieves a classification from Diamond, is it still anchored at Root? i.e.
**Root;** d__Archaea; p__Proteobacteria

wwood:
I assume you are talking about reads that are hit by diamond. If so, yes - see https://github.com/wwood/graftM/commit/a30dd80a8a8dab832a195f66dbf5b5c12653351a#diff-35a1ac6a033bf152d4fa462f86a0ecc4R459

geronimp:
Good, good.

PR #111, opened 2015-09-09
create: add --search_hmm_files

wwood:
I'm thinking there should be no reason to hack CONTENTS.json files any longer? Hoping graftM create is up for the challenge now.


geronimp:
All this code looks okay to me, but one more things needs to be added to make graftM create "self-sufficient", and that is unaligned sequences. It should be simple to do, but currently when you provide unaligned sequences they aren't added the GraftM package, and no CONTENTS.json entry is made. In the branch I'm working on now, "unaligned_sequence_database" is a required. Could you add the code that copies the unaligned sequences into the package? Just since you're working on this bit of the code.

wwood:
Should unaligned sequences be included if they are duplicates?

geronimp:
No, duplicates are worthless. Why is that?

PR #127, opened 2015-11-12
Refpkg tree path

wwood:
Hey a pull request after #126 just to make things a little less hacky. Untested still since the base branch doesn't pass the tests. A simple method though.


wwood:
Just added a new commit so the name of the method is consistent with other method names in the class

geronimp:
Thanks, I'll merge because the graftM update branch is still failing. I'll deal with it.

PR #137, opened 2015-11-17
Rename decorate to tree

wwood:
further to #136


geronimp:
are these good to merge?

wwood:
On your say so..

PR #143, opened 2015-11-20
Prerequisite check considers all cases for FastTree

geronimp:


wwood:
a good start, but if it is fasttree not FastTreeMP, then the actual command line will fail, no?

geronimp:
*Still going*

geronimp:
Should be GG now. I've passes all tests.

geronimp:
Am I good to merge this?

wwood:
Thanks. This is an annoying one. What are your thoughts on making the 
object that gives the paths back a singleton one, liking the logging? 
Seems quite messy passing this as a parameter everywhere. Probly a good 
idea t change the name too from PrerequisiteChecker as the class has 
changed quite fundamentally.

geronimp:
Hopefully all good now.

wwood:
Seems there's a merge conflict

geronimp:
Indeed. Probably minor. I can merge now and see?

wwood:
merge master to your branch, not the other way around, yeh?

geronimp:
ja.

geronimp:
very minor indeed

wwood:
file missing

graftm/external_program_suite.py
```diff
@@ -29,20 +34,39 @@ def check_prerequisites(program_list):
...
+Arbitrarily selecting %s" % (' '.join(program_list), check[0]))
+                    
+                for key, item in ExternalProgramSuite\
+                                        .programs_to_possibilities\
+                                        .iteritems():
+                    if program in item:
+                        program = key
+                
+                if key == "FastTreeMP":
+                    self.fasttree = check[0]
```
wwood: this doesn't feel right as a lonesome string, but eh for now.

wwood:
would be good to add a test for the fasttree detection (test_external_program_suite.py ?), and comment Creat#init but other than that looks good - I assume you've nosetested it.

geronimp:
This test is fairly basic for now. Seeing as it is not possible to load/unload module via python I'm not really sure how to expand this. What do you think?

wwood:
maybe the equivalent of
```
mkdir /tmp/tempdir
ln -s `which FastTree` /tmp/tempdir
setenv PATH `sed 's/:/\n/g' |grep -iv fasttree |tr '\n' ':'`
do the test
setenv PATH old_path
```
?

geronimp:
Sorry - missing the point of the soft link in the first two lines

wwood:
think I mean
```
ln -s `which FastTree` /tmp/tempdir/fasttree
```
better?

geronimp:
How about now?

wwood:
sorry do you mind merging master to your branch first, so don't have to inspect the changes twice? thanks.

geronimp:
Okay done, It passes nosetests too.

test/test_external_program_suite.py
```diff
@@ -0,0 +1,85 @@
...
+        
+            fasttree_original_path=which("FastTree")
+            fasttree_possibilities = {'FastTreeMP':'fasttreeMP',
+                                      'FastTree':'fasttree'}
+        elif which("fasttree"):
+            fasttree_original_path=which("fasttree")
+            fasttree_possibilities = {'fasttreeMP':'FastTreeMP',
+                                      'fasttree':'FastTree'}
+        else:
+            raise NoFastTreeExecError("No version of fasttree is installed that graftM can recognise!")
```
wwood: might be easier to just use
```
self.fail("No version of fasttree is installed that graftM can recognise!")
```
rather than defining a special class.

graftm/external_program_suite.py
```diff
@@ -19,8 +19,13 @@ class PrerequisiteChecker:
                             'ktImportText': 'krona',
                             'FastTreeMP': 'FastTree'}
     
-    @staticmethod
-    def check_prerequisites(program_list):
+    programs_to_possibilities = {
+                                 "FastTreeMP": ["FastTreeMP",
+                                                "fasttree",
+                                                "fasttreeMP"]
```
wwood: `FastTree` too?

wwood:
just some minor things, otherwise good to go. If it is OK with you I'll do the merging though, so let me know about the changes I suggest. ta.

geronimp:
Changes added - you're free to merge

wwood:
It seems `graftm/prerequisite_checker.py` has been deleted from this branch (in 9bc660f081a37fbafb). I'm not sure what the go is with this, and whether changes have been made since so I can't just restore the file.

thanks.

geronimp:
I believe prerequisite checker was replaced by external_program_suite.py


This branch crashes because of a stray PrerequisiteChecker import that I forgot to remove. I'll push this change when I confirm nosetests passes after the fix.


Joel

geronimp:
This checks out of nosetests now.

wwood:
closing since this is merged already it seems

PR #145, opened 2015-11-28
18s hmm fix

wwood:
fix for the issue chris had on friday.

also made the dependencies >= instead of ==. I don't think there is a good solution, but other notable python packages e.g. skbio seem to leave them open ended, so let's do that.

Shall we call this 0.9.2? I'll push it out on your say so.
ta


geronimp:
Have you tried nosetesting this? Seems to hand on the 8th test or so. Wondering if you noticed this.

wwood:
Seems to work fine for me, can you ctrl-c during the hang for a stack 
trace or something?
ta

geronimp:
No dice I'm afraid:

uqjboyd2@keating:~/04_gits/git_graftM/test$ nosetests -x
.......^C
----------------------------------------------------------------------
Ran 8 tests in 1013.799s

OK


I'll try an narrow down which it is.

geronimp:
My problem, it would seem. Works fine now. I've merged and we should be good to call this 0.9.2

PR #146, opened 2015-11-30
Can now update contents file within a graftM package

geronimp:
Can now update CONTENTS file of GraftM package with a little more ease.


wwood:
I'm not sure I quite follow. To me, specifying --graftm_package indicates that I want to use all the data in the graftm package to create a new package, except that extra parameters I provide e.g. --search_hmm_files are used in preference to those in the graftm package. So then updating a package is just a convenience way to set specify commandline parameters like --rerooted-tree.  Make sense? The problem with having a separate method for updating is that there is effectively duplicated code. Also, a test or two would be good. ta

geronimp:
I'll come back to this - It isn't urgent for now right?

wwood:
sure, no problem.

PR #150, opened 2015-12-02
create: don't read bootstraps as taxonomy in annotated trees

wwood:
solves some of Caitlin's problems, but not all.


wwood:
better? I'm not sure there is a strong convention on this, it was just in the code I grabbed from stackoverflow.

PR #154, opened 2015-12-07
add failing test for input sequence with gaps

wwood:
Hey,

This isn't for merging, it needs more work - this is just proof of fail, but I don't have time right now. Wanna give it a go?

### DIFF
--- a/test/test_graft.py
+++ b/test/test_graft.py
@@ -1394,5 +1394,62 @@ FADPHLVFDFREPKMCIAKATLRQFMPAGERDPTLPPH
                     self.assertEqual(128, e.returncode)
                     raise e
 
+    def test_protein_input_for_pplacer(self):
+        testing_read = '''>Methanoflorens_stordalmirensis_v4.3.2_01362 Methyl-coenzyme M reductase subunit alpha
+MATEKTQKMFLEAMKKKFAEDPTSNKTTYKREGWTQSKDKREFQEWGAKIAKDRGIPAYN
+VNVHLGGMTLGQRQLMPYNVSGTDVMCEGDDLHYVNNPAMQQMWDEIRRTVIVGLDTAHE
+TLTRRLGKEVTPETINGYLEALNHTMPGAAIVQEHMVETHPALVEDCFVKVFTGDDDLAA
+EIDKPFLVDINKLFPKAQAEQLKKAIGKTLWQGVHIPTIVSRTCDGGNTSRWSAMQIGMS
+FIAAYNMCAGEAAVADLAFAAKHASLVEMANFLPARRARGPNEPGGLSFGFMADMIQTDR
+IYPEDPVKSSLEVVAAGCMLYDQIWLGSYMSGGVGFTQYATAAYTDNILDDYSYYGNDYA
+KKYGEDGKAPSTMDVVNDLGTEVTLYGIEQYEKYPTTLEDHFGGSQRATVLAAASGVTVA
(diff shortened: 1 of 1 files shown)


geronimp:
sure

geronimp:
Do you want this to fail? For now I'll throw a warning

wwood:
I was hoping it would work, not fail. But if a warning is much easier in the short term then ok.

geronimp:
Yep, I'll make it work but warn the user they've input a gapped seq in

wwood:
ah ok cool, good idea

geronimp:
There are two ways we could do this:
One is that we accept the input alignment to be re-searched the hmm, which means if it hits the alignment length will match the pplacer tree, or we could skip the search step and attempt to place pre-aligned sequences automatically.

I'm thinking implement the former and provide the latter as a flag. Thoughts?

wwood:
Hmm, hadn't thought of that. I generally run across this when I'm providing a sequence that is part of a different multiple sequence alignment, and I intend simply for the gaps to be stripped before searching. Have you had need to do the latter in your experience?

geronimp:
Well I can imagine it would be useful if you had an alignment handy. But the searching/alignment isnt the real bottleneck of graftM anyway, so it might be too much trouble for now to implement. Maybe leave it till someone asks I say. 

After playing around with it, it seems that you cant hmmsearch/hmmalign pre-aligned sequences and get the same length alignment as the HMM length, causing pplacer to fail. Looks like I'll have to strip the '-' out like you say

I like this a whole lot though. If you're using hmmaligned sequences you may have masked out some possible important aa's along the way, which would be lost when '-'s are stripped out - know what I mean?

But a simple warning for the user should suffice I think.

wwood:
ok

geronimp:
This patch will still fail in the instance that the amino acid pipeline is being run, and that the first alignment gap ('-') in the file is >100 lines down. I dont want to read in the whole file for timing/memory reasons.

Basically this is because hmmsearch works on alignments, meaning this case goes undetected till the pplacer step. It would be very difficult to implement a workaround in that part of the code. This is not an issue for the nucleotide search step because nhmmer fails if an '-' in in the sequence, meaning the exception can be caught there and then and worked around.


I think it will need some work to find a workaround. Right now I've got other things to do. In the meantime. This branch passes nosetests.

wwood:
So if I'm understanding this, hmmsearch works but then puts the `-` characters back in the sequence, and then because there is no lower case letters in the amino acids the fasta alignment the column doesn't get masked out?

If so, maybe the best workaround is to parse the stockholm file directly rather than relying on seqmagick, and hopefully then be able to detect the difference between a regular gap and a gap from the input sequence?

Also I'm not seeing a patch.

geronimp:
Strange, I pushed and it worked fine.. Give me a second

And no - that's not quite what I meant. Hang on I'll re-push

wwood:
OK got the patch now, thanks. I think I understand what the approach is in that patch (and from your comment), but is what I said right?

geronimp:
The '-'s arent re-introduced. they are removed if they are detected and run through the usual pipeline. Perhaps we can discuss in person tomorrow morning.


wwood:
what's the go here?

geronimp:
I think the patch is good to go; as for your q:

It may be possible yes - I can look into this if you like. Avoiding seqmagick may be a + for speed as well.

Joel

wwood:
which patch? the original patch here is just a failing test

PR #156, opened 2015-12-10
remove obsolete graftm packages to avoid confusion

geronimp:
They aren't necessary anymore.


wwood:
github page is having a hard time with this, but if this is just deletion of the gpkgs folder then go for it.

PR #157, opened 2015-12-11
actually run the tests with setup.py test

wwood:
Currently this runs the tests, although they fail atm.


wwood:
retried in #159

PR #158, opened 2015-12-14
_remove_gaps_patch

geronimp:


wwood:
Hey just took the easy root and redid the merge and force pushed that as master since there was still code left from the merge, so no need for these extra patches now. Thanks anyway.

PR #159, opened 2015-12-14
actually run the tests with setup.py test

wwood:
same as before rebased to the new master, is this ok?


geronimp:
looks okay. Unfortunately these changes missed the latest docker/pip release - but nothing too important hey

PR #161, opened 2015-12-14
catch bady placed root and use alternative re-rooting method.

geronimp:
not for merging - Working on one of Caitlins bugs

### DIFF
--- a/graftm/rerooter.py
+++ b/graftm/rerooter.py
@@ -108,8 +108,8 @@ class Rerooter:
         elif len(children) == 2:
             raise BinaryTreeError("Tree is already binary. Something has \
 -mislead me here!")
-        elif len(children) < 2:
-            raise MalformedTreeError("Has only one child from the root!") # Impossible?
+        else:
+            raise MalformedTreeError("Input tree is malformed")
             
 ###############################################################################
 ###############################################################################
--- a/graftm/rerooter.py
+++ b/graftm/rerooter.py
@@ -32,7 +32,7 @@ from skbio import TreeNode
 
 ################################ - Classes - ##################################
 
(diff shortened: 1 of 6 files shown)


wwood:
OK, thanks. No need to send a pull request unless you are after a review
though.

geronimp:
Yes I'm after a review sorry, I wrote the message before my email

geronimp:
I still want to merge this because I made some minor improvements to the re-rooting script. Let me know what you think

wwood:
OK, sure. Probly should cherry-pick out the improvements only and include on a separate branch then tho.

geronimp:
All of the changes are good, except for the mid point rerooting which is now useless. I'll just push another commit with that part removed if you dont mind.

wwood:
A test or two might be in order.

geronimp:
True

geronimp:
Hey, I've added a test that (VERY) simply tests that the re-rooting is doing what it should do. Do you think any other test would help out here?

- J

graftm/rerooter.py
```diff
@@ -86,12 +95,21 @@ def reroot(self, tree):
...
+                            length=0)
             return tree
-        else:
+        
+        elif len(children) == 2:
             raise BinaryTreeError("Tree is already binary. Something has \
-mislead me here!")
-
+-mislead me here!")
+        elif len(children) < 2:
```
wwood: just else is better here I think

wwood:
I think that's it,

wwood:
ok go ahread and merge then, I take it the tests pass

geronimp:
had that prerequisite error - I'll remove it in this branch as well and confirm that tests pass

PR #165, opened 2016-01-11
add failing test_input_sequence_near_duplication

wwood:
Seems in the output of this test there is only 1 sequence in the combined alignment, and only one sequence in the .jplace file, when there should be 2. Not sure what the go is.. any ideas?

### DIFF
--- a/test/test_graft.py
+++ b/test/test_graft.py
@@ -1185,7 +1185,53 @@ FISDQMVGHKLGEFAPTRTFRGHAKSDKKGRR
                 self.assertEqual(expected, open(os.path.join(tmp, os.path.basename(fasta.name)[:-3], "%s_hits.fa" % os.path.basename(fasta.name)[:-3])).read())
         
 
-
+    def test_input_sequence_near_duplication(self):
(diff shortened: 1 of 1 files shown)


geronimp:
If the sequences are identical they are dereplicated and then placed into the tree. This makes the placement step a little faster - especially when dealing with amplicon datasets. 

It does that issue of misleading names for files again though - some more thought into the naming of these files may be needed.

In any case within the output for this run you should still find that the number of sequences reported in the basic stats, and the output subdirectory should have both aligned sequences and the read taxonomy file should also have two lines.

If you find that this is not the case just post here and I'll look into it.

wwood:
Ah ok, I see.

Thing is, people look at these files as results files, not intermediate files. Should we remake these alignment and jplace files so that they have the correct number of sequences in them i.e. undereplicate them? They are misleading as is I think, atm they are intermediate files so shouldn't even appear in the output folder I'd say.

geronimp:
Yep, hence my second line. I'll leave this branch open to remind me.

wwood:
hey how come this got closed? Has the test been included in another branch with cherry-pick or otherwise?

wwood:
jplace rereplicated in #201, but rereplication of alignment still not implemented.

PR #168, opened 2016-01-14
Filter minimum update

wwood:
The default 95bp trimming seems quite stringent to me, for matching reads. Have you done any science tuning this parameter?


geronimp:
Better than what I did

PR #170, opened 2016-01-18
geronimp-length_filter_bugfix

geronimp:
Just fixing a bug that occurs when you specify a length cutoff of 0


wwood:
This is already fixed in my branch I think, do you mind having a browse of my pull requests please?

PR #171, opened 2016-01-19
geronimp-slash_error patch

geronimp:
Patch to the /2 /1 ending error. I also added another test which tests for another bug I found when working with /2 /1 endings too. Some changes to test_hmmer.py that accommodate for the different order which the read merging outputs reads now - no effect here on output of graftM graft.

### DIFF
--- a/graftm/clusterer.py
+++ b/graftm/clusterer.py
@@ -3,6 +3,7 @@ from graftm.sequence_io import SequenceIO
 from graftm.orfm import OrfM
 import logging
 import os
+import re
 
 class Clusterer:
 
@@ -10,7 +11,7 @@ class Clusterer:
         self.clust = Deduplicator()
         self.seqio = SequenceIO()
         self.seq_library = {}
-
+        self.slash_ending_regex = re.compile('.*/[12]$')
         self.orfm_regex = OrfM.regular_expression()
 
 
@@ -47,7 +48,9 @@ class Clusterer:
                 placed_alignment_base = placed_alignment_file.replace('_clustered.fa', '')
             output_annotations[placed_alignment_base] = {}
             for rep_read_name, rep_read_taxonomy in cluster_classifications.iteritems():
-
+                slashendingregex = self.slash_ending_regex.match(rep_read_name)
+                if slashendingregex:
+                    rep_read_name=rep_read_name[:-2]
                 if reverse_pipe:
                     orfm_regex = OrfM.regular_expression()
                     clusters={(orfm_regex.match(key).groups(0)[0] if orfm_regex.match(key) else key): item for key, item in clusters.iteritems()}
@@ -68,12 +71,14 @@ class Clusterer:
             list of strings, each a path to input fasta files to be clustered.
         reverse_pipe : bool
             True/False, whether the reverse reads pipeline is being followed.
+
         Returns
         -------
         output_fasta_list : list
             list of strings, each a path to the output fasta file to which
             clusters were written to.
         '''
+        
         output_fasta_list = []
         for input_fasta in input_fasta_list:
             output_path  = input_fasta.replace('_hits.aln.fa', '_clustered.fa')
@@ -91,7 +96,11 @@ class Clusterer:
                                         output_path
                                         ) # Choose the first sequence to write to file as representative (all the same anyway)
             for cluster in clusters:
-                cluster_dict[cluster[0].name]=cluster # assign the cluster to the dictionary
+                if self.slash_ending_regex.match(cluster[0].name):
+                    name = cluster[0].name[:-2]
+                else:
+                    name = cluster[0].name                    
+                cluster_dict[name]=cluster # assign the cluster to the dictionary
             self.seq_library[output_path]= cluster_dict
 
             output_fasta_list.append(output_path)
--- a/graftm/hmmer.py
+++ b/graftm/hmmer.py
@@ -5,6 +5,7 @@ import itertools
 import logging
 import tempfile
 import subprocess
+import re
 
 from Bio import SeqIO
 from collections import OrderedDict
@@ -229,37 +230,39 @@ class Hmmer:
         -------
         Nothing - output files are known.
         '''
-
+        slash_ending_regex = re.compile('.*/[12]$')
         orfm_regex = OrfM.regular_expression()
-        def remove_orfm_end(records):
+        
+        def clean_read_headers(records):
 
             new_dict = {}
 
             for key, record in records.iteritems():
+                slashregex = slash_ending_regex.match(key)
+                if slashregex:
+                    key = key[:-2]
+                
                 orfmregex = orfm_regex.match(key)
                 if orfmregex:
-                    new_dict[orfmregex.groups(0)[0]] = record
-                else:
-                    new_dict[key] = record
+                    key = orfmregex.groups(0)[0] 
+                    
+                new_dict[key] = record
+            
             return new_dict
 
 
         for idx, (forward_path, reverse_path) in enumerate(zip(forward_aln_list, reverse_aln_list)):
+            
             output_path = outputs[idx]
             logging.info('Merging pair %s, %s' % (os.path.basename(forward_path), os.path.basename(reverse_path)))
-            forward_reads = SeqIO.parse(forward_path, 'fasta')
-            reverse_reads = remove_orfm_end(SeqIO.to_dict(SeqIO.parse(reverse_path, 'fasta')))
-
+            forward_reads = clean_read_headers(SeqIO.to_dict(SeqIO.parse(forward_path, 'fasta')))
+            reverse_reads = clean_read_headers(SeqIO.to_dict(SeqIO.parse(reverse_path, 'fasta')))
+            
             with open(output_path, 'w') as out:
-                for forward_record in forward_reads:
-                    regex_match = orfm_regex.match(forward_record.id)
-                    if regex_match:
-                        id = regex_match.groups(0)[0]
-                    else:
-                        id = forward_record.id
+                for key, forward_record in forward_reads.iteritems():
                     forward_sequence = str(forward_record.seq)
                     try:
-                        reverse_sequence = str(reverse_reads[id].seq)
+                        reverse_sequence = str(reverse_reads[key].seq)
                         new_seq = ''
                         if len(forward_sequence) == len(reverse_sequence):
                             for f, r in zip(forward_sequence, reverse_sequence):
@@ -274,16 +277,17 @@ class Hmmer:
                                         new_seq += f
                                 else:
                                     new_seq += '-'
+                        
                         else:
                             logging.error('Alignments do not match')
                             raise Exception('Merging alignments failed: Alignments do not match')
                         out.write('>%s\n' % forward_record.id)
                         out.write('%s\n' % (new_seq))
-                        del reverse_reads[id]
+                        del reverse_reads[key]
                     except:
-
                         out.write('>%s\n' % forward_record.id)
                         out.write('%s\n' % (forward_sequence))
+                
                 for record_id, record in reverse_reads.iteritems():
                     out.write('>%s\n' % record_id)
                     out.write('%s\n' % (str(record.seq)))
--- a/graftm/run.py
+++ b/graftm/run.py
@@ -232,7 +232,7 @@ class Run:
        
         # If merge reads is specified, check that there are reverse reads to merge with
         if self.args.merge_reads and not hasattr(self.args, 'reverse'):
-            logging.error("--merge requires --reverse to be specified")
+            logging.error("--merge_reads requires --reverse to be specified")
             exit(1)
 
         # Set the output directory if not specified and create that directory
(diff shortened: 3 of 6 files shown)


graftm/unpack_sequences.py
```diff
@@ -1,9 +1,11 @@
 import logging
 import subprocess
-from signal import signal, SIGPIPE, SIG_DFL
 import os
 import itertools
+import re
```
wwood: are there extra imports here?

graftm/hmmer.py
```diff
@@ -229,37 +230,39 @@ def merge_forev_aln(self, forward_aln_list, reverse_aln_list, outputs):
         -------
         Nothing - output files are known.
         '''
-
+        slash_ending_regex = re.compile('.*/[12]$')
```
wwood: technically this is the same as `re.compile('/[12]$')`

graftm/clusterer.py
```diff
@@ -47,7 +48,9 @@ def uncluster_annotations(self, input_annotations, reverse_pipe):
                 placed_alignment_base = placed_alignment_file.replace('_clustered.fa', '')
             output_annotations[placed_alignment_base] = {}
             for rep_read_name, rep_read_taxonomy in cluster_classifications.iteritems():
-
+                slashendingregex = self.slash_ending_regex.match(rep_read_name)
```
wwood: matching it in several places seems quite repetitive, might be best to detect it once and then set a variable in the unpack_sequences object, and then refer to that? Not 100% sure on this.

graftm/unpack_sequences.py
geronimp: Yes... Perhaps I did intend on matching the /1 /2 endings here after all.

graftm/clusterer.py
geronimp: That is a more sustainable option yeah, I just got lazy. I'll be a good coder and fix this.

graftm/hmmer.py
geronimp: Hmm I'm not sure:
```
In [3]: header
Out[3]: '>FCC0WM1ACXX:2:2208:12709:74426#GTCCAGAA/1'

In [4]: re.compile('/[12]$').match(header)

In [5]: re.compile('.*/[12]$').match(header)
Out[5]: <_sre.SRE_Match at 0x3fd6cc8>
```
Am I doing something wrong?

wwood: Ah right I'd forgotten. Python uses the (I believe non-standard) 
approach that match has to apply to the beginning of the string, where 
another method (search?) does not.

geronimp:
There is some pretty ugly backwards incompatibility here. Hopefully for the better.

geronimp:
Clusterer is fully documented on my branch so I'm not sure what you're referring to here.

wwood:
It isn't documented looking at the diff here in 'files changed'. Strange the way github is showing those comments though, something is odd

geronimp:
Am I good to merge?

wwood:
something is off between 'your branch' (unsure what that means), and what you are presenting here. There is no comments here for instance:
```
class Clusterer:

    def __init__(self, slash_endings):
        self.clust = Deduplicator()
        self.seqio = SequenceIO()
        self.seq_library = {}
        self.slash_endings = slash_endings
        self.orfm_regex = OrfM.regular_expression()
``

**Comment by geronimp** [comment]
Oops, I didn't realise which part you were referring to. How is the above?

**Comment by geronimp** [comment]
Hey just bumping this. This is the nosetest error that's failing I believe

File: graftm/clusterer.py [hunk 4]
```diff
@@ -47,7 +62,8 @@ def uncluster_annotations(self, input_annotations, reverse_pipe):
                 placed_alignment_base = placed_alignment_file.replace('_clustered.fa', '')
             output_annotations[placed_alignment_base] = {}
             for rep_read_name, rep_read_taxonomy in cluster_classifications.iteritems():
-
+                if merge_reads:
+                    rep_read_name=rep_read_name[:-2]
```
wwood: doesn't you can merge reads that don't have the slash though, right?

geronimp: Thanks for pointing this out. I did have self.slash_endings there to check this, but it appears I didnt put it where it was actually needed, which was silly.

wwood: You've run unit tests on this branch yeh?

geronimp: Well I thought I did originally, but it fails when normal reads are provided so obviously not. I'm rerunning them now.

geronimp: uqjboyd2@mrca002:~/04_gits/git_graftM/test$ nosetests -x
....................................................................................................................
----------------------------------------------------------------------
Ran 116 tests in 445.509s

OK

geronimp:
Hey, I've addressed your comments and it passes nosetests.

test/test_guess_sequence_type.py
```diff
@@ -24,18 +24,27 @@
...
-        self.assertEqual('nucleotide', urr._guess_sequence_type_from_string('A'*300+'E'*999)) #only look at the first 300bp
-        self.assertEqual('nucleotide', urr._guess_sequence_type_from_string('a'*10+'T'*89)) #lowercase
+        
+        test_read=""">HWI-ST1243:121:D1AF9ACXX:8:1101:12684:12444
+GTCAACAACCCCGCAATGCAACAGATGTGGGATGAGATCAGACGTACAGTTATCGTCGGTCTCGACCAGGCCCACGAGACGCTGACCAGAAGACTCGGTAAGGAAGTTACCCCTGAGACCATCAACGGCTATCTTGAGGCGTTGAACCAC"""
+        with tempfile.NamedTemporaryFile(suffix='.fa') as fasta:
+            fasta.write(test_read)
+            fasta.flush()
+
+            urr = UnpackRawReads(fasta.name)
```
wwood: Did you meant to test what sequence type comes back when you provide a file? If so there is no assertEqual for this. Also, this method tests _guess_sequence_type_from_string, I'd make a separate test method for testing instantiating it using a file.

graftm/unpack_sequences.py
```diff
@@ -35,6 +37,10 @@ class UnexpectedFileFormatException(Exception): pass
     
     def __init__(self, read_file):
         self.read_file   = read_file
+        self.slash_ending_regex = re.compile('.*/[12]$')
```
wwood: I don't think this should be an instance variable, as it is only used internally to this class. Instead, simply define it as a regular variable in `sequence_type`?

graftm/unpack_sequences.py
```diff
@@ -77,9 +83,17 @@ def sequence_type(self):
...
                                                 )
-            _, seq = tuple(first_seq.strip().split('\n'))
+            header, seq = tuple(first_seq.strip().split('\n'))
             self.type = self._guess_sequence_type_from_string(seq)
             logging.debug("Detected sequence type as %s" % self.type)
+            
+            slashendingregex = self.slash_ending_regex.match(header)
+            if slashendingregex:
+                logging.debug("Detected '/1' '/2' style headers.")
+                self.slash_endings = True
```
wwood: I'd prefer boolean variables to be named like `has_slash_endings` or similar. `slash_endings` reads like the variable actually holds the strings `/1` or something.

graftm/hmmer.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import tempfile
 import subprocess
+import re
```
wwood: unneeded import?

wwood:
ok to merge after those minor comments.

for bonus points, maybe good to merge this 'squashed'?
https://github.com/blog/2141-squash-your-commits

geronimp:
Cool, thanks


graftm/clusterer.py
```diff
@@ -7,11 +7,11 @@
 
 class Clusterer:
 
-    def __init__(self, slash_endings):
+    def __init__(self, has_slash_endings):
         '''
         Parameters
         ----------
-        slash_endings : bool
+        has_slash_endings : bool
```
wwood: actually I'm confused. By making has_slash_endings an instance variable of Clusterer, are you assuming that all read sets either have or all read sets do not have, the slash endings? Might be good to test for this, too

Shouldn't has_slash_endings be derived from the Unpack, not the Clusterer?

geronimp: You're right, it does assume. I'll take a look at fixing this.

geronimp:
Hey Ben, I've addressed your commend above. Unfortunately it requires some changes to run.py but hopefully they all make sense. This branch passes all nosetests, but I needed to change one test slightly to comply with the changes  made to a function in Hmmer. 

Hopefully we can merge after this final change though.

Joel

graftm/clusterer.py
```diff
@@ -73,7 +66,7 @@ def uncluster_annotations(self, input_annotations, reverse_pipe,
 
         return output_annotations
 
-    def cluster(self, input_fasta_list, reverse_pipe):
+    def cluster(self, unpacks_list, reverse_pipe):
```
wwood: the comments need updating here

graftm/hmmer.py
```diff
@@ -238,7 +237,7 @@ def merge_forev_aln(self, has_slash_endings,
 
         orfm_regex = OrfM.regular_expression()
         
-        def clean_read_headers(records):
+        def clean_read_headers(records, has_slash_endings):
```
wwood: this needs commenting too

graftm/unpack_sequences.py
wwood: please remove

graftm/db_search_results.py
```diff
@@ -19,6 +19,6 @@ def search_objects(self): # Return SequenceSearchResult object with parameters d
     def hit_count(self): # Return the number of hits found
         return self.hit_count  
     
-    def slash_endings(self): # Return True if sequence headers end in /1 or /2, False if not
-        return self.slash_endings
+    def has_slash_endings(self): # Return True if sequence headers end in /1 or /2, False if not
```
wwood: I think this method is superfluous, because `has_slash_endings` is accessible from any code that has this object. So I would remove it.

wwood: well, that may entail some code change, but it just means there is only 1 way to do this, which makes things less confusing. And less code is better, mostly.

graftm/hmmer.py
```diff
@@ -210,16 +210,21 @@ def hmmsearch(self, output_path, input_path, unpack, seq_type, threads, cutoff,
...
         Parameters
         ----------
-        aln_list : array
-            List of the forward and reverse alignments for each of the runs
-            given to graftM. **MUST** be the following pattern:
-            [forward_run1, reverse_run1, forward_run2, reverse_run2 ...]
+        has_slash_endings : bool
+            True/False: whether the sequence headers contain the old style 
+            '/1' and '/2' headers to denote forward and reverse reads.
+        for_aln_list : array
```
wwood: these comments are no longer correct - please update.

graftm/hmmer.py
```diff
@@ -231,35 +236,38 @@ def merge_forev_aln(self, forward_aln_list, reverse_aln_list, outputs):
         '''
 
         orfm_regex = OrfM.regular_expression()
-        def remove_orfm_end(records):
+        
+        def clean_read_headers(records, has_slash_endings):
```
wwood: please add a comment

graftm/run.py
```diff
@@ -381,7 +381,8 @@ def graft(self):
                                                         filter_minimum
                                                         )
                     if aln_result:
-                        seqs_list.append(hit_aligned_reads)
+                        setattr(unpack, 'alignment_path', hit_aligned_reads)
```
wwood: I think adding an `alignment_path` instance variable to an unpack object which ostensibly an object for representing where a read comes from is making things a bit too confusing, yeh?

The problem is some information flow? I'm a bit lost so this could be off base, but does it make sense to invent a new class that holds an alignment path and whether they have slash endings, or something along those lines?

wwood:
Hey,

I didn't quite make it through sorry, got stuck on this bit:

`setattr(unpack, 'alignment_path', hit_aligned_reads)`

I think that doing something like that will make it harder to refactor 
`Run` in the long term. Maybe a class that holds the results of a search 
step e.g. unpack, alignment, has_slash_endings ?

geronimp:
Okay, thank you.
I've left that last comment because I need to clarify exactly what you mean.

graftm/unpack_sequences.py
geronimp: Everything here is used I believe

wwood: ah ok sorry

wwood:
Hey, we should merge this, where are we at?

geronimp:
Those conflicts look just awful. I think just by looking at your that comments I've left unanswered here there is some work left to do. Do you think, considering how far behind this branch is, it may be worth seeing how easy it is to re-fix this problem on the current master?

wwood:
Potentially, yeh.



-- 
Ben Woodcroft
DECRA Recipient & Post-doctoral Research Fellow

Australian Centre for Ecogenomics, UQ, Australia
P: +617 336 54957  W: http://www.ecogenomic.org

geronimp:
Okay I'll take a look at that then, thank you

PR #172, opened 2016-01-19
Geronimp jplace improvements

geronimp:
Hey here is the .jplace file patch. 

It is synced with the slash error branch so please have a look at that one first just (https://github.com/geronimp/graftM/pull/171) just so you're aware of all changes that are within this branch.

Thanks,
Joel

### DIFF
--- a/graftm/clusterer.py
+++ b/graftm/clusterer.py
@@ -7,15 +7,16 @@ import re
 
 class Clusterer:
 
-    def __init__(self):
+    def __init__(self, slash_endings):
         self.clust = Deduplicator()
         self.seqio = SequenceIO()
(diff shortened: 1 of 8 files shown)


geronimp:
Instance variable check has been added.

wwood:
Sorry to be a bit narky, but in future would you mind please keeping fixes for different bugs in different branches? Reviewing fixes for multiple different bugs all in follow-on pull requests makes it difficult to understand your intent when reading the code. If they are in different branches we can merge fixes in the order that they pass review, rather than in the order that they are proposed. I won't bother reviewing this for now then. Thanks.

geronimp:
That was my bad, apologies.

PR #181, opened 2016-02-10
Neater, more understandable, and hopefully bugfree code for GraftM tr…

geronimp:
…ee. This update also includes a Taxonomy module which will eventually be implemented across GraftM. Currently taxonomy is read in at multiple places in the GraftM code, making it less modular and les flexible. In the future, this Taxonomy module will used to read in taxonomy in run.py in the first instance and passed into child modules from there. The tree decorator module was completely redesigned allowing for other decoration and tree analysis functions to be added more easily. Decoration of new sequences now works without error as far as I can detect. There are no more stacked species clusters as there were before thanks to the simplified code.

### DIFF
--- a/graftm/decorator.py
+++ b/graftm/decorator.py
@@ -1,10 +1,39 @@
-
+#!/usr/bin/env python
+###############################################################################
+#                                                                             #
+# This program is free software: you can redistribute it and/or modify        #
+# it under the terms of the GNU General Public License as published by        #
+# the Free Software Foundation, either version 3 of the License, or           #
+# (at your option) any later version.                                         #
+#                                                                             #
+# This program is distributed in the hope that it will be useful,             #
+# but WITHOUT ANY WARRANTY; without even the implied warranty of              #
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                #
+# GNU General Public License for more details.                                #
+#                                                                             #
+# You should have received a copy of the GNU General Public License           #
+# along with this program. If not, see <http://www.gnu.org/licenses/>.        #
+#                                                                             #
+###############################################################################
+__author__ = "Joel Boyd, Ben Woodcroft"
+__copyright__ = "Copyright 2015"
+__credits__ = ["Joel Boyd"]
+__license__ = "GPL3"
+__version__ = "0.0.1"
+__maintainer__ = "Joel Boyd"
+__email__ = "joel.boyd near uq.net.au"
+__status__ = "Development"
+###############################################################################
 # System imports
 import logging
 from skbio import TreeNode
 from graftm.reannotator import Reannotator
 from graftm.tree_decorator import TreeDecorator
 
+# Local imports
+################################################################################
+################################################################################
+################################################################################
 class Decorator:
     '''re-root a tree and decorate it using old taxonomy, for the graftM 
     decorate pipeline.
@@ -14,7 +43,7 @@ class Decorator:
     decorated by this class is WILL be overwritten. Of course, bootstrap values 
     will remain untouched.'''
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs):     
         '''
         Parameters
         ----------
@@ -23,7 +52,7 @@ class Decorator:
             retroot the tree tree provided to tree
         old_tree_path: str
             Path to the file containing the tree to be re-rooted. This tree will
-            be rerooted at the same position as the tree porovided to the 
+            be rerooted at the same position as the tree provided to the 
             reference_tree
         '''
         reference_tree_path = kwargs.pop('reference_tree_path', None)
@@ -48,8 +77,7 @@ class Decorator:
         self.tree = reannotator._reroot_tree_by_old_root(self.reference_tree, 
                                                          self.tree)
         
-    def main(self, taxonomy, output_tree, output_tax, no_unique_tax,
-             decorate, seqinfo):
+    def main(self, taxonomy, output_tree, output_tax, no_unique_tax, decorate):
         '''Decorate and if necessary, re-root the tree. If an old reference tree
         is provided it is assumed that re-rooting is desired
         
@@ -64,20 +92,28 @@ class Decorator:
             Path to file to which the decorated tree will be written to.
         output_tax: str
             Path to file to which the decorated taxonomy will be written to.
+            A boolean False if no output taxonomy file was specified. In this
+            case no taxonomy is written to file.
         seqinfo: str
             path to taxtastic seqinfo file to be used with taxonomy to annotate
-            the tree.'''        
+            the tree.
+        '''        
+        
         # Reroot
         if self.reference_tree:
             self._reroot()
+            
         # Decorate
-        td =  TreeDecorator(self.tree,
-                            taxonomy,
-                            seqinfo)
         if decorate:
-            td.decorate(output_tree, output_tax, no_unique_tax)
+            td =  TreeDecorator(self.tree,
+                                no_unique_tax)
+            td.taxonomy_partition(taxonomy)
+            
+        if output_tax:
+            td.write_taxonomy(output_tax)
         else:
-            logging.info("Writing tree to file: %s" % (output_tree))
+            logging.info("Writing tree to file: %s" % output_tree)
+            self.tree = td.return_tree()
             self.tree.write(
                             output_tree,
                             format = "newick"   
--- a/graftm/run.py
+++ b/graftm/run.py
@@ -21,6 +21,7 @@ from graftm.timeit import Timer
 from graftm.clusterer import Clusterer
 from graftm.decorator import Decorator
 from graftm.external_program_suite import ExternalProgramSuite
+from graftm.taxonomy import Taxonomy
 from biom.util import biom_open
 
 T=Timer()
@@ -44,7 +45,7 @@ class Run:
         self.setattributes(self.args)
 
     def setattributes(self, args):
-        
+        self.t = Taxonomy()
         self.hk = HouseKeeping()
         self.s = Stats_And_Summary()
         if args.subparser_name == 'graft':
@@ -649,12 +650,21 @@ class Run:
                 # any of the provided paths.
                 gpkg = GraftMPackage.acquire(self.args.graftm_package)
                 if not self.args.rooted_tree: self.args.rooted_tree = gpkg.reference_package_tree_path()
+
                 if not self.args.input_greengenes_taxonomy:
                     if not self.args.input_taxtastic_seqinfo:
                         self.args.input_taxtastic_seqinfo = gpkg.taxtastic_seqinfo_path()
                     if not self.args.input_taxtastic_taxonomy:
                         self.args.input_taxtastic_taxonomy = gpkg.taxtastic_taxonomy_path()
-                        
+               
+            if self.args.input_greengenes_taxonomy:
+                self.t.read_format_greengenes(open(self.args.input_greengenes_taxonomy))
+            elif self.args.input_taxtastic_seqinfo and self.args.input_taxtastic_taxonomy:
+                self.t.read_format_taxtastic(open(self.args.input_taxtastic_seqinfo),
+                                             open(self.args.input_taxtastic_taxonomy))
+            else:
+                logging.error("Some form of taxonomic description must be provided.")
+            
             if self.args.rooted_tree:
                 if self.args.unrooted_tree:
                     logging.error("Both a rooted tree and an un-rooted tree were provided, so it's unclear what you are asking GraftM to do. \
@@ -680,25 +690,14 @@ If you're unsure see graftM tree -h")
                 logging.error("Some tree(s) must be provided, either a rooted tree or both an unrooted tree and a reference tree")
                 exit(1)
                 
-            if self.args.output_taxonomy is None and self.args.output_tree is None:
-                logging.error("Either an output tree or taxonomy must be provided")
+            if self.args.output_tree is None:
+                logging.error("An output tree must be provided!")
                 exit(1)
-            if self.args.input_greengenes_taxonomy:
-                if self.args.input_taxtastic_seqinfo or self.args.input_taxtastic_taxonomy:
-                    logging.error("Both taxtastic and greengenes taxonomy were provided, so its unclear what taxonomy you want graftM to decorate with")
-                    exit(1)
-                logging.debug("Using input GreenGenes style taxonomy file")
-                dec.main(self.args.input_greengenes_taxonomy, 
-                         self.args.output_tree, self.args.output_taxonomy, 
-                         self.args.no_unique_tax, self.args.decorate, None) 
-            elif self.args.input_taxtastic_seqinfo and self.args.input_taxtastic_taxonomy:
-                logging.debug("Using input taxtastic style taxonomy/seqinfo")
-                dec.main(self.args.input_taxtastic_taxonomy, self.args.output_tree, 
-                         self.args.output_taxonomy, self.args.no_unique_tax, 
-                         self.args.decorate, self.args.input_taxtastic_seqinfo)
-            else:
-                logging.error("Either a taxtastic taxonomy or seqinfo file was provided. GraftM cannot continue without both.")
-                exit(1)
+            
+            dec.main(self.t, 
+                     self.args.output_tree, self.args.output_taxonomy, 
+                     self.args.no_unique_tax, self.args.decorate) 
+            
         else:
             raise Exception("Unexpected subparser name %s" % self.args.subparser_name)
             
--- a/graftm/taxonomy.py
+++ b/graftm/taxonomy.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python
+###############################################################################
+#                                                                             #
+# This program is free software: you can redistribute it and/or modify        #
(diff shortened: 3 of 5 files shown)


geronimp:
I just realised that I have yet to implement taxtastic taxonomy imports into Taxonomy.py. Please dont review until I have - I'll keep you posted.

geronimp:
Actually go ahead if that's okay. I'll implement taxtastic taxonomy imports when I merge in the improvements to run.py. 

This code passes all tests except the slash error one.

graftm/decorator.py
```diff
@@ -1,10 +1,39 @@
...
+# This program is distributed in the hope that it will be useful,             #
+# but WITHOUT ANY WARRANTY; without even the implied warranty of              #
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                #
+# GNU General Public License for more details.                                #
+#                                                                             #
+# You should have received a copy of the GNU General Public License           #
+# along with this program. If not, see <http://www.gnu.org/licenses/>.        #
+#                                                                             #
+###############################################################################
+__author__ = "Joel Boyd, Ben Woodcroft"
```
wwood: Hey I think including the copyright notice is a good idea, but I think the extra `__author__` is a bit too much to maintain - WDYT?

graftm/decorator.py
```diff
@@ -64,20 +92,28 @@ def main(self, taxonomy, output_tree, output_tax, no_unique_tax,
             Path to file to which the decorated tree will be written to.
         output_tax: str
             Path to file to which the decorated taxonomy will be written to.
+            A boolean False if no output taxonomy file was specified. In this
+            case no taxonomy is written to file.
         seqinfo: str
             path to taxtastic seqinfo file to be used with taxonomy to annotate
-            the tree.'''        
+            the tree.
```
wwood: can be None?

graftm/run.py
```diff
@@ -44,7 +45,7 @@ def __init__(self, args):
         self.setattributes(self.args)
 
     def setattributes(self, args):
-        
+        self.t = Taxonomy()
```
wwood: why make this an instance variable?

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+                    continue
+                else:
+                    ammended_split_taxonomy = \
+                                taxa_taxonomy_list + \
+                                [''] * (self.max - len(taxa_taxonomy_list))
+                self.taxonomy[taxa_id] = ammended_split_taxonomy
+        except:
+            logging.error("Controlled exit from taxonomy.Taxonomy \
+ _ammend_taxonomy_gaps")    
+            ProgrammingError("Error when trying to interpret taxonomy input")
```
wwood: need `raise` ?

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+                
+                if len(taxa_taxonomy_list) == self.max:
+                    continue
+                else:
+                    ammended_split_taxonomy = \
+                                taxa_taxonomy_list + \
+                                [''] * (self.max - len(taxa_taxonomy_list))
+                self.taxonomy[taxa_id] = ammended_split_taxonomy
+        except:
+            logging.error("Controlled exit from taxonomy.Taxonomy \
```
wwood: Seems pretty uncontrolled to me. Not sure a log entry is needed here.

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+        ----------
+        taxonomy_io : Open greengenes taxonomy file
+        '''
+        logging.debug("Reading in greengenes formatted taxonomy")
+        try:
+            for entry in taxonomy_io:
+                entry_id, entry_taxonomy_string = entry.strip()\
+                                                       .split(self.TAB)
+                
+                if self.GG_SEP_0 in entry_taxonomy_string:
```
wwood: Isn't it easier to just split on `;` and then strip?

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+        Parameters
+        ----------
+        taxonomy_io : Open greengenes taxonomy file
+        '''
+        logging.debug("Reading in greengenes formatted taxonomy")
+        try:
+            for entry in taxonomy_io:
+                entry_id, entry_taxonomy_string = entry.strip()\
+                                                       .split(self.TAB)
+
```
wwood: Does python croak if there is >2 entries in the split?

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+                                entry_taxonomy_string.split(self.GG_SEP_0)                
+                else:
+                    split_entry_taxonomy = \
+                                 entry_taxonomy_string.split(self.GG_SEP_1)
+                
+                self.taxonomy[entry_id] = split_entry_taxonomy
+            self._ammend_taxonomy_gaps()
+            self.id_set = set(self.taxonomy.keys())
+        except:
+            logging.error("Controlled exit from taxonomy.Taxonomy \
```
wwood: As above, not really necessary.

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+read_format_greengenes")
+            raise MalformedGreenGenesFormattedFile("The GreenGenes taxonomy file \
```
wwood: Be good to include the line that was malformed, helps the user fix their GG file.

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+                               'g__',
+                               's__'])
+
+    def __init__(self, tree, no_unique_names):
         '''
-        Set up empty lists and open and import tree object, and load the
-        taxonomy into a hash
+        Parameters
+        ----------
+        tree: skbio TreeNode obj
```
wwood: just `skbio.tree.TreeNode` I think.

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+            http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.html#skbio.tree.TreeNode
```
wwood: `The tree to be decorated`

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+        no_unique_names : bool
+            True of False, whether names will be made non-redundant if
```
wwood: I think it is more understandable to avoid double negatives, so better to use `generate_unique_names` so to make unique names.

But more generally, I think this style of class where there is a bunch of methods to setup the parameters for a single call (here `taxonomy_partition`) isn't optimal because there are several ways to provide parameters - via the instance variable, in the init and as parameters directly to taxonomy_partition. THere's also 2 types of methods, those for before decoration and those for after.

I reckon a better setup would be to Have a taxonomy decorator class, and then a separated DecoratedTree class, which has the write_taxonomy methods and etc. 

Make sense?

graftm/decorator.py
geronimp: Yes, A good example: I also neglected to add your information below, which was poor form. Sorry about that.

I'll boot the __author__ stuff.

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
-                    else:
-                        logging.warning("No taxonomy found for species %s!" % (tax_name))
-                        tax_string = "Unknown"
-                else:
-                    tax_string = '; '.join(tax_list)
-                output_line = "%s\t%s\n" % (tax_name, tax_string)
-                out.write(output_line)
-    
-    def _get_tax_index(self, ancestry):
+        taxonomy: TaxonomyObject
```
wwood: just `Taxonomy` not `TaxonomyObject` no?

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+                 
+        Returns
+        -------
+        skbio TreeNode object
         '''
-        Iterate through ancestor nodes to find the current taxonomy index (i.e.
-        the rank that should be assigned to the current node.
+        self.max_taxonomy_index=taxonomy.max
```
wwood: Why set this as an instance variable? It is only used in this method

graftm/taxonomy.py
geronimp: Yes

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+        consistency_check = 1
+        seen_taxonomy_rank = {}
```
wwood: I make a habit of naming hashes to be key_to_value, so in this case maybe `taxonomy_to_next_unique`, maybe you want to follow this, maybe not

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
-                
-                if ancestor.name:
-                    split_node_name = ancestor.name.split(':')
-                    if len(split_node_name) == 2:
-                        ancestor_tax_list.append(split_node_name[1])
-                    elif len(split_node_name) == 1:
-                        pass
+        for node in self.tree.preorder():    
+            tax_string_array = []
+            for rank in range(0, self.max_taxonomy_index):
```
wwood: I think the `0, ` is superfluous

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+                node_classifications = []
```
wwood: I'd use the keys from an OrderedDict, so that you don't have to repeatedly create new sets with `set(array)`

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+                    ammended_split_taxonomy = \
+                                taxa_taxonomy_list + \
+                                [''] * (self.max - len(taxa_taxonomy_list))
+                self.taxonomy[taxa_id] = ammended_split_taxonomy
+        except:
+            logging.error("Controlled exit from taxonomy.Taxonomy \
+ _ammend_taxonomy_gaps")    
+            ProgrammingError("Error when trying to interpret taxonomy input")
+    
+    def check(self, taxa_id):
```
wwood: Maybe better name this method `__get__` so that you can then do `if tip_name in taxonomy:`

graftm/tree_decorator.py
geronimp: Makes sense yeah. I wont call it taxonomy decorator though, just decorator. Just because not just tax can be used here.

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+        
+        if taxa_id in self.taxonomy:
+            taxonomy_string = GG_SEP_0.join(self.taxonomy[taxa_id])
+        else:
+            raise MissingTaxonomy("Taxa id provided did not have any \
+any corresponding entry in the taxonomy description.")
+            
+        return taxonomy_string
+            
+    def read_format_greengenes(self, taxonomy_io):
```
wwood: I think it might be better to make these methods `@staticmethod`s which return a new Taxonomy instance so you can `taxonomy = Taxonomy.read_format_greengenes(io)` yeh?

wwood: Why not just grab the code from `graftm/greengenes_taxonomy.py` ?

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
+        taxonomy: TaxonomyObject
+                 
+        Returns
+        -------
+        skbio TreeNode object
         '''
-        Iterate through ancestor nodes to find the current taxonomy index (i.e.
-        the rank that should be assigned to the current node.
+        self.max_taxonomy_index=taxonomy.max
+        consistency_check = 1
```
wwood: this variable name doesn't make sense to me.

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+        else:
+            return False
+        
+    def entry_ids(self):
+        '''
+        Return a list of all taxonomy entries within the self.taxnomy dict
+        '''
+        return self.taxonomy.keys()
+    
+    def rank(self, taxa_id, rank):
```
wwood: To me this method name reads like "get me the rank of Acidobacteria". I wonder if instead you can just run `taxonomy[taxon][rank]` instead - this method is not complex enough to exist imo

graftm/taxonomy.py
```diff
@@ -0,0 +1,176 @@
...
+
+    TAB='\t'
+    COM=','
+    GG_SEP_0='; '
+    GG_SEP_1=';'
+    
+    def __init__(self):
+        self.taxonomy = {}
+        
+    def _ammend_taxonomy_gaps(self):
```
wwood: amend with 1 m. Why is it necessary to have all the same length?

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
-                
-                if resolution < MAX_RESOLUTION:
-                    new_tax_list = [] 
+                            tax_string_array.append(classification) 
+                            seen_taxonomy_rank[classification] = 1
+                            
+            tax_string_array = [tax for tax in tax_string_array
+                                if tax not in self.GREENGENES_PREFIXES]
+            
+            if any(tax_string_array):
```
wwood: do you mind adding some comments here, I'm getting confused.

graftm/tree_decorator.py
```diff
@@ -1,309 +1,187 @@
...
-                    node.name = node_name
-                    logging.debug("Renamed node %s" % node_name)
-                    self.encountered_nodes[node]=index
+                tax_string_array = tax_string_array[index:]
+                
+                if len(tax_string_array) >0:
+                    node.name =  Taxonomy.GG_SEP_0.join(tax_string_array)
+                    node.index = len(tax_string_array)
+        
+    def return_tree(self):
```
wwood: why not just return the tree when it has been through `taxonomy_partition`?

graftm/run.py
geronimp: It isn't necessary, I just kept it together with HouseKeeping() and Stats_And_Summary(). Do you think this should not be an instance variable (Does it cause issues?)

graftm/taxonomy.py
geronimp: Yes, added

graftm/taxonomy.py
geronimp: Originally left them in so the error was captured in the .log file output, because I'm not sure if Exceptions are written to the file.

I've taken these out.

graftm/taxonomy.py
geronimp: Yep. Added.

graftm/taxonomy.py
geronimp: .. But there shouldn't be if the file is formatted correctly.

graftm/taxonomy.py
geronimp: added

graftm/tree_decorator.py
geronimp: My mistake

graftm/run.py
geronimp: I recall I was going to use the Taxonomy class elsewhere in the GraftM code (explained in the commit description)

graftm/tree_decorator.py
geronimp: fair enough.

graftm/tree_decorator.py
geronimp: I'm not sure what difference that would make in this instance sorry.

I've implemented a workaround so sets are not used at all. They were kinda used in a dirty way here anyways.

graftm/taxonomy.py
geronimp: Thanks for pointing this out. I was wondering how to do this and now I do.

graftm/taxonomy.py
geronimp: Will probably do that ^

graftm/tree_decorator.py
geronimp: changed to consistent_annotation which is a boolean. True if the node has consistent annotation or False if not...
Is that better?

graftm/tree_decorator.py
geronimp: If more functions are added to tree_decorator (which I intended). This allows me to call functions on the tree (specified by the user and interpreted through decorator.main) until I need the end result i.e. when the tree is written to file. That way I wont be playing ping pong with the tree whenever I need to decorate it with something else.

graftm/tree_decorator.py
geronimp: will be in next commit.

graftm/taxonomy.py
geronimp: You're right. I've modified this.

graftm/taxonomy.py
geronimp: So that when we search for the phylogenetic rank of a specific tip (as you mentioned above) the taxonomy[tip_name][rank] will not choke.

graftm/tree_decorator.py
geronimp: Thats a good idea

graftm/tree_decorator.py
geronimp: In light of my comments below, are you okay with keeping it like this?

graftm/tree_decorator.py
geronimp: actually because its needed in write_taxonomy

PR #189, opened 2016-04-27
Geronimp investigating pplacer issue

geronimp:
pretty prints,  parses properly, passes tests

### DIFF
--- a/graftm/pplacer.py
+++ b/graftm/pplacer.py
@@ -40,7 +40,8 @@ class Pplacer:
                 for record in alignments: # For each record in the read list
                     record.id = record.id + '_' + str(file_number) # append the unique identifier to the record id
                 SeqIO.write(alignments, output, "fasta") # And write the reads to the file
-                alias_hash[str(file_number)] = {'output_path': os.path.join(os.path.dirname(alignment_file),'placements.jplace')}
+                alias_hash[str(file_number)] = {'output_path': os.path.join(os.path.dirname(alignment_file),'placements.jplace'),
+                                                'place': []}
                 file_number += 1
         return alias_hash
 
@@ -50,22 +51,24 @@ class Pplacer:
         # Parse the placements based on unique identifies appended to the end
         # of each read
         for placement in original_jplace['placements']: # for each placement
-            for alias_idx in alias_hash.keys(): # For each alias, append to the 'place' list each read that identifier
-                alias_placements_list = []
-                nm_list = []
-                p = placement['p']
-                for nm in placement['nm']:
-                    read_alias_idx = nm[0].split('_')[-1] # Split the alias 
-                                        # index out of the read name, which 
-                                        # corresponds to the input file from 
-                                        # which the read originated.
-                    if read_alias_idx == alias_idx:
-                        nm_list.append(nm)
-                if any(nm_list):
-                    placement_hash = {'p': p,
-                                      'nm': nm_list}
-                    alias_placements_list.append(placement_hash)                
-                alias_hash[alias_idx]['place'] = alias_placements_list
+            
+            alias_placements_list = []
+            nm_dict = {}
+            p = placement['p']
+            for nm in placement['nm']:
+                read_alias_idx = nm[0].split('_')[-1] # Split the alias 
+                                    # index out of the read name, which 
+                                    # corresponds to the input file from 
+                                    # which the read originated.
+                
+                if read_alias_idx not in nm_dict:
+                    nm_dict[read_alias_idx] = [nm]
+                else:
+                    nm_dict[read_alias_idx].append(nm)
+            for alias_idx, nm_list in nm_dict.iteritems():
+                placement_hash = {'p': p,
+                                  'nm': nm_list}                
+                alias_hash[alias_idx]['place'].append(placement_hash)
         return alias_hash
     
     def write_jplace(self, original_jplace, alias_hash):
--- a/graftm/pplacer.py
+++ b/graftm/pplacer.py
@@ -44,16 +44,12 @@ class Pplacer:
                 file_number += 1
         return alias_hash
 
-    def jplace_split(self, jplace_file, alias_hash):
+    def jplace_split(self, original_jplace, alias_hash):
         ## Split the jplace file into their respective directories
-
-        # Load the placement file
-        placement_file = json.load(open(jplace_file))
         
         # Parse the placements based on unique identifies appended to the end
         # of each read
-        for placement in placement_file['placements']: # for each placement
-            
+        for placement in original_jplace['placements']: # for each placement
             for alias_idx in alias_hash.keys(): # For each alias, append to the 'place' list each read that identifier
                 alias_placements_list = []
                 nm_list = []
@@ -68,23 +64,21 @@ class Pplacer:
                 if any(nm_list):
                     placement_hash = {'p': p,
                                       'nm': nm_list}
-                    alias_placements_list.append(placement_hash)
-                    
+                    alias_placements_list.append(placement_hash)                
                 alias_hash[alias_idx]['place'] = alias_placements_list
-
+        return alias_hash
+    
+    def write_jplace(self, original_jplace, alias_hash):
         # Write the jplace file to their respective file paths.
-        jplace_path_list = []
         for alias_idx in alias_hash.keys():
-            output = {'fields'     : placement_file['fields'],
-                      'version'    : placement_file['version'],
-                      'tree'       : placement_file['tree'],
+            output = {'fields'     : original_jplace['fields'],
+                      'version'    : original_jplace['version'],
+                      'tree'       : original_jplace['tree'],
                       'placements' : alias_hash[alias_idx]['place'],
-                      'metadata'   : placement_file['metadata']}
+                      'metadata'   : original_jplace['metadata']}
             with open(alias_hash[alias_idx]['output_path'], 'w') as output_io:
                 json.dump(output, output_io, ensure_ascii=False, indent=3, separators=(',', ': '))
-            jplace_path_list.append(alias_hash[alias_idx]['output_path'])
-        return jplace_path_list
-    
+
     @T.timeit
     def place(self, reverse_pipe, seqs_list, resolve_placements, files, args,
               slash_endings, tax_descr):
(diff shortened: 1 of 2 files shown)


geronimp:
except the slash error, of course because we're on master

geronimp:
Do you mind talking a look at this one also?

graftm/pplacer.py
```diff
@@ -69,7 +76,7 @@ def jplace_split(self, jplace_file, alias_hash):
                       'placements' : alias_hash[alias]['place'],
                       'metadata'   : placement_file['metadata']}
             with open(alias_hash[alias]['output_path'], 'w') as output_path:
-                json.dump(output, output_path, ensure_ascii=False)
+                json.dump(output, output_path, ensure_ascii=False, indent=3, separators=(',', ': '))
```
wwood: I'd call that `output_io` not `output_path` as it not a path.

graftm/pplacer.py
```diff
@@ -50,16 +50,23 @@ def jplace_split(self, jplace_file, alias_hash):
...
-
+                nm_list = []
+                p = placement['p']
+                for nm in placement['nm']:
+                    if nm[0].split('_')[-1] == alias:
+                        nm_list.append(nm)
+                if any(nm_list):
+                    hash = {'p': p,
+                            'nm': nm_list}
+                    alias_hash[alias]['place'].append(hash)
```
wwood: this is probably fine but slightly dangerous, because you are changing the same object that is being iterated over. I'd make a separate variable and build upon that.

graftm/pplacer.py
```diff
@@ -50,16 +50,23 @@ def jplace_split(self, jplace_file, alias_hash):
 
         # Load the placement file
         placement_file = json.load(open(jplace_file))
-
+        
         # Parse the placements based on unique identifies appended to the end
         # of each read
         for placement in placement_file['placements']: # for each placement
             hash = {} # create an empty hash
```
wwood: this statement does nothing because `hash` is always re-instantiated later anyway

graftm/pplacer.py
```diff
@@ -50,16 +50,23 @@ def jplace_split(self, jplace_file, alias_hash):
...
             for alias in alias_hash: # For each alias, append to the 'place' list each read that identifier
-                hash = {'p': placement['p'],
-                        'nm': [nm for nm in placement['nm'] if nm[0].split('_')[-1] == alias]}
-                alias_hash[alias]['place'].append(hash)
-
+                nm_list = []
+                p = placement['p']
+                for nm in placement['nm']:
+                    if nm[0].split('_')[-1] == alias:
```
wwood: why are you `split`ing here? Might be best to add a comment because it implies a long-range code coupling, I'm guessing.

wwood:
The method `jplace_split` does 2 things: re-add the aliased read names, and write the jplace to file(s). I would split it into the two methods.

When you've split then you can add lots of tests to make sure that it is splitting the jplaces correctly, without having to `extern.run "graftM graft ..."`. Instead you can just provide it with the json hash and the aliases. I think we should add tests because this code is rarely encountered by users, so bug reports are likely not to come up very often.

Thanks for the small, self-contained bug fixes. Much easier to review compared to a long-diverged branch.

geronimp:
Okay, no worries. Thanks for the comments.

graftm/pplacer.py
geronimp: If > 1 file is given to GraftM it "barcodes" each sequence so that the file of origin (i.e. the alias) for each sequence is always known. This way we only need to run pplacer once, as opposed to running it for every input file. I'll add the comment in.

geronimp:
Hey, sorry for big-ish changes, but I think this branch is good to go.

geronimp:
passes tests, except slash ending error.

graftm/pplacer.py
```diff
@@ -40,39 +40,49 @@ def alignment_merger(self, alignment_files, output_alignment_path):
...
                     record.id = record.id + '_' + str(file_number) # append the unique identifier to the record id
                 SeqIO.write(alignments, output, "fasta") # And write the reads to the file
-                alias_hash[str(file_number)] = {'output_path': os.path.join(os.path.dirname(alignment_file),'placements.jplace'),
+                alias_hash[str(file_number)] = {'output_path': os.path.join(os.path.dirname(alignment_file), 'placements.jplace'),
                                                 'place': []}
                 file_number += 1
         return alias_hash
 
-    def jplace_split(self, jplace_file, alias_hash):
+    def jplace_split(self, original_jplace, alias_hash):
```
wwood: be good to add a little doco here. `alias_hash` in particular is non-obvious

wwood: especially since this method is not 'pure'
https://en.wikipedia.org/wiki/Pure_function

test/test_json_placement_split.py
```diff
@@ -0,0 +1,154 @@
...
+
+import unittest
+import os
+import sys
+from graftm.pplacer import Pplacer
+
+
+sys.path = [os.path.join(os.path.dirname(os.path.realpath(__file__)),'..')]+sys.path
+
+class HmmsearcherTests(unittest.TestCase):
```
wwood: class name should be just `Tests`

graftm/pplacer.py
wwood: might also be good to mention that it doesn't deal with `"n"` style jplace format files (I presume pplacer never returns these?)

test/test_json_placement_split.py
```diff
@@ -0,0 +1,154 @@
...
+                      {"invocation":
+                        "pplacer -c test_16S.gpkg\/test_16S.gpkg.refpkg\/ GraftM_output\/combined_alignment.aln.fa"
+                      }
+                    }
+        
+        pplacer = Pplacer("refpkg_decoy")
+        
+        output_alias_hash = pplacer.jplace_split(test_json, input_alias_hash)
+        
+        for alias_idx, alias_placement_entry in output_alias_hash.iteritems():
```
wwood: apparently in python3 `iteritems` is no longer a thing. Might be good to start getting in the habit of using `items()` instead.

graftm/pplacer.py
geronimp: Probably worth a mention. I added doco to this in a separate branch I made a PR for recently. I can add it there if you think its worth mentioning

geronimp: Maybe just a check for v3 jplace file would be sufficient

test/test_json_placement_split.py
geronimp: Good to know, thanks

graftm/pplacer.py
wwood: v3 can contain `n` style though right?

wwood: so I should not bother reviewing this PR and just go off the other one?

test/test_json_placement_split.py
```diff
@@ -0,0 +1,154 @@
...
+            
+            placement_dict = alias_placement_entry['place']
+            self.assertEqual(len(placement_dict), 1)
+            self.assertEqual([x[0] for x in placement_dict[0]['p']],
```
wwood: I find this assert statement a bit confusing because I'm a bit slow. Why not just something like this?
```
self.assertEqual( ['p__Proteobacteria', 'k__Bacteria', 'p__Proteobacteria'],
    placement_dict[0]['p'][0])
```
While it doesn't matter practically, it is conventional to write tests like 
```
assertEqual(expected, observed)
```
not the reverse order.

Also confusing because the 'expected value' is not called 'expected', and is defined so far away in terms of line number. I try not to be fancy when writing tests e.g. loops when there is only 1 element, because it confuses things.

graftm/pplacer.py
geronimp: If you like. It has the same changes, so you havent lost ground. I made an extra branch because I was fixing another, unrelated issue and needed the changes in this branch.

test/test_json_placement_split.py
```diff
@@ -0,0 +1,154 @@
...
+                             placement)
+        
+    def test_basic_split_with_different_placements(self):
+        input_alias_hash = {'0':{'place':[]}, '1':{'place':[]}}
```
wwood: can you add a quick test case for a situation where there is actually an alias please?

graftm/pplacer.py
geronimp: But I did add comments to some functions in that branch also.

wwood: ok well I'll go off the other branch now then. Do you mind making sure what I comment on here is addressed though? Don't want to have to cross-reference. ta

wwood:
closed in favour of #201 which includes these commits.

wwood:
hey let me me know when you are finito with these changes Joel and then I'll start on the other branch.

geronimp:
Sure, I'll push the changes to the other branch then

test/test_json_placement_split.py
geronimp: Not sure what you mean here. This is formatted exactly like it should be in the code.

wwood: I mean add another test, where the `[]` actually has some content i.e. there is some aliases. Make sense?

geronimp: The alias refers to the key in this dictionary, each number corresponding to a different input file. The 'place' list is the jplace placement info for each read in that file (like this):
```
 {"nm":[["test_read1_0", 1] , ["test_read2_0", 1]]
"p": [["p__Proteobacteria", 0.107586583111, 1, 0.970420466541, -614.032176075, 0.22226616471],
 .... and so on 
]}
```

geronimp: So I'm a little confused, sorry

wwood: I think it is me that is confused! I guess I don't understand the function of the '0' and '1'.

test/test_json_placement_split.py
wwood: or better yet just test the whole thing
```python
assertEqual({'0': {'place': ['p__Proteobacteria', 'k__Bacteria', 'p__Proteobacteria']},
    '1': {'place': ['p__Proteobacteria', 'k__Bacteria', 'p__Proteobacteria']}},
   placement_dict)
```

test/test_json_placement_split.py
geronimp: 0 = input file 1
1 = input file 2

thats what the "alias" is. I guess you could call in index, but its just a numbering system to distinguish the origin of the reads that is simple to parse.

test/test_json_placement_split.py
geronimp: Okay this has been done on https://github.com/geronimp/graftM/pull/201

wwood: not quite, but good enough

Issue #192, opened 2016-05-02
taxonomy not extracted correctly from trees with branch lengths?


wwood:
e.g. `/srv/projects/abisko/annotation/12_dsr_aterrible12/original_package_creation/sulfitereductase.ben.tree`

I'll fix this one.

wwood:
this actually works fine, the error was all me, I've added a test just in case.

PR #193, opened 2016-05-03
taxonomy_extractor: add test for branch lengths, ref #192.

wwood:


geronimp:
You can merge if you like, although adding more tests for things like bootstrap values might be useful also.

wwood:
there's already 2 test for bootstraps in there, is there some other problem?

PR #194, opened 2016-05-03
Conform to skbio 0.4.2.

wwood:
at long last. I think there is a bug in skbio where it doesn't accept open IO objects to TreeNode.read, but we can workaround that by just providing the path to the file, rather than opening it first.


graftm/create.py
```diff
@@ -747,8 +747,7 @@ def main(self, **kwargs):
                 tre_file_tempfile = tempfile.NamedTemporaryFile(suffix='.tree', prefix='graftm')
                 tre_file = tre_file_tempfile.name
                 with tempfile.NamedTemporaryFile(suffix='.tree', prefix='graftm') as f:
-                    tree.write(f)
-                    f.flush()
+                    tree.write(f.name)
```
geronimp: Why do we remove flush?

geronimp:
Okay, just that one question and you're good to merge

graftm/create.py
wwood: we are giving `TreeNode.write` a string path, not an IO object. So it does open, write, and close for us. The context (if that is the word) opens f as an IO object, but actually we don't need that, we just want the tempfile. But writing it like this is convenient so that the tempfile is deleted at the end of the context.

geronimp: alright, thanks

wwood:
merged

PR #195, opened 2016-05-03
create: test package works after creation

wwood:
I got sick of create going through but then later getting nasty surprises..

I'd prefer to do something along the lines of
```
Graft().graft(...)
```
But unfortunately `Run` is tightly coupled to argparse (`args` is an instance variable). If we make a new `Graft` class then we might need to also define a `Graft.DEFAULT_PARAMETERS` so that defaults can be specified in `bin/graftM` and in the method definition of `Graft.graft` without repetition (repetition might mean they become out of sync, which would be bad). ..future work.

### DIFF
--- a/graftm/create.py
+++ b/graftm/create.py
@@ -821,6 +821,18 @@ in the final GraftM package. If you are sure these sequences are correct, turn o
         logging.info("Cleaning up")
         self._cleanup(self.the_trash)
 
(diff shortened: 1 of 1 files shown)


geronimp:
ok, feel free to merge... Maybe raise this as an issue so its not forgotten?

PR #197, opened 2016-05-03
archive: new mode.

wwood:
diamond dbs are a bit too chunky which makes it hard to put singlem on pip packaged with particular gpkgs. So here's a new mode which tar.gz's a gpkg, minus the diamond db, and then reconstitutes it.

### DIFF
--- a/graftm/archive.py
+++ b/graftm/archive.py
@@ -1,0 +1,114 @@
+import tarfile
+import logging
+import itertools
+import os
+import tempdir
+import shutil
+
+from graftm_package import GraftMPackage
+
+class ArchiveDefaultOptions:
+    force=False
+
+class Archive:
+    def create(self, input_package_path, output_package_path, **kwargs):
+        """Create an archived GraftM package
+
+        Parameters
+        ----------
+        input_package: GraftMPackage
+            Package to be archived
+        output_pacakge_path: str
+            output package path
+        kwargs:
+            force: bool
+                overwrite an existing directory
+
+        Returns
+        -------
+        nothing"""
+        force = kwargs.pop('force',ArchiveDefaultOptions.force)
+        if len(kwargs) > 0:
+            raise Exception("Unexpected arguments detected: %s" % kwargs)
+
+        logging.info("Archiving GraftM package '%s' as '%s'" % (input_package_path, output_package_path))
+        gpkg = GraftMPackage.acquire(input_package_path)
+        if gpkg.version != 3:
+            raise Exception("Archiving GraftM packages only works with format 3 packages")
+
+        self._setup_output(output_package_path, force)
+
+        def add_file(tar, path):
+            # Put the gpkg folder itself in the archive so as not to tar bomb.
+            tar.add(path,
+                    os.path.join(os.path.basename(os.path.abspath(gpkg._base_directory)),
+                                 os.path.basename(path)))
+
+        logging.debug("Compressing contents for archive")
+        with tarfile.open(output_package_path, 'w:gz') as tar:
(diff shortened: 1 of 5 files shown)


graftm/archive.py
```diff
@@ -0,0 +1,114 @@
...
+            Package to be archived
+        output_pacakge_path: str
+            output package path
+        kwargs:
+            force: bool
+                overwrite an existing directory
+
+        Returns
+        -------
+        nothing"""
```
geronimp: I'm not sure if this is needed

graftm/archive.py
```diff
@@ -0,0 +1,114 @@
...
+        force = kwargs.pop('force',ArchiveDefaultOptions.force)
```
geronimp: Why not make this a global variable if you use it in extract as well?

graftm/archive.py
```diff
@@ -0,0 +1,114 @@
...
+
+        logging.debug("Compressing contents for archive")
+        with tarfile.open(output_package_path, 'w:gz') as tar:
+            for path in itertools.chain([gpkg.contents_file_path(),
+                                         gpkg.alignment_hmm_path(),
+                                         gpkg.reference_package_path(),
+                                         gpkg.unaligned_sequence_database_path()],
+                                        gpkg.search_hmm_paths()):
+                logging.debug("Compressing '%s'" % path)
+                add_file(tar, path)
```
geronimp: Why the separate function to do one thing?

graftm/archive.py
wwood: ok

graftm/run.py
```diff
@@ -707,6 +708,67 @@ def main(self):
...
+            |          |                     |                 `+------+
+            |          |                     |
+            |          |                     |
+            |          |                     |
+            |          |                     |
+            |          |                     |
+            `-.        |                  _.-'
+               `-.     |           __..--'
+                  `-.  |      __.-'
+                     `-|__.--'
```
geronimp: Very nice, I like it a lot

graftm/archive.py
wwood: I think you mean instance variable not global variable? Either way, I'd prefer not to make things have more scope than necessary.

graftm/archive.py
wwood: historical reasons, will fix

test/test_archive.py
```diff
@@ -0,0 +1,55 @@
...
+        with tempdir.in_tempdir():
+            archiver = Archive()
+            ingpkg = os.path.join(path_to_data, 'mcrA.gpkg')
+            arc = 'compressed.gpkg.tar.gz'
+            out = 'reconstituted.gpkg'
+            archiver.create(ingpkg, arc)
+            archiver.extract(arc, out)
+            self.assertEquals(
+                ['CONTENTS.json', 'mcrA.faa', 'mcrA.faa.dmnd', 'mcrA.hmm', 'mcrA.refpkg'],
+                sorted(os.listdir(out)))
```
geronimp: You could use:
```
from filecmp import dircmp
gpkg_compare = dircmp(ingpkg, out)
assertEquals(gpkg_compare.diff_files, [])
```
Which has the advantage of actually checking that the files are identical, rather than just the file names.

geronimp:
I'm happy otherwise. Feel free to merge when you're ready

test/test_archive.py
wwood: can't do that because there is extra files in the ingpkg, but I'll check for contents equality with filecmp.cmp. Thanks for that tip.

wwood:
merged via a72a9fc

PR #201, opened 2016-05-04
Geronimp investigating pplacer issue rereplicate jplace files

geronimp:
Passes all tests except slash error. This addresses the issue pointed out in the unmerged branch https://github.com/geronimp/graftM/pull/165 where the output placement files are not filled out with replicates (i.e. only de-replicated sequences are in the placement files).

I've fixed this issue, added a test and also removed the file aliases appended to the end of each sequence name in the jplace files. See commit comments below for specifics of the changes.

Joel

### DIFF
--- a/graftm/pplacer.py
+++ b/graftm/pplacer.py
@@ -121,14 +121,6 @@ class Pplacer:
         # Run pplacer on merged file
         jplace = self.pplacer(files.jplace_output_path(), args.output_directory, files.comb_aln_fa(), args.threads)
         logging.info("Placements finished")
-
-        # Split the original jplace file
-        # and write split jplaces to separate file directories
-        jplace_json = json.load(open(jplace))
-        alias_hash_with_placements = self.jplace_split(jplace_json, 
-                                                       alias_hash)
-        self.write_jplace(jplace_json, 
-                          alias_hash_with_placements)
         
         #Read the json of refpkg
         logging.info("Reading classifications")
@@ -162,6 +154,14 @@ class Pplacer:
                 trusted_placements[base_file]={}
                 for read, entry in classifications[str(idx)].iteritems():
                     trusted_placements[base_file][read] = entry['placement']
+        
+        # Split the original jplace file
+        # and write split jplaces to separate file directories
+        jplace_json = json.load(open(jplace))
+        alias_hash_with_placements = self.jplace_split(jplace_json, 
+                                                       alias_hash)
+        self.write_jplace(jplace_json, 
+                          alias_hash_with_placements)
         
         return trusted_placements
 
--- a/graftm/pplacer.py
+++ b/graftm/pplacer.py
@@ -44,8 +44,33 @@ class Pplacer:
                                                 'place': []}
                 file_number += 1
         return alias_hash
-
-    def jplace_split(self, original_jplace, alias_hash, clusterer):
+    
+    def convert_cluster_dict_keys_to_aliases(self, cluster_dict, alias_hash):
+        '''
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


geronimp:
latest wwood comments commit addressing some comments in the  https://github.com/geronimp/graftM/pull/189 pull request

graftm/pplacer.py
```diff
@@ -40,42 +40,103 @@ def alignment_merger(self, alignment_files, output_alignment_path):
...
+        abundance (re-replicates?) into the placement file so analyses can be 
+        done using the placement files. 
+        
+        Parameters
+        ----------
+        original_jplace : dict (json)
+            json .jplace file from the pplacer step.
+        alias_hash : dict
+            Stores information on each input read file given to GraftM, the
+            corresponding reads found within each file, and their taxonomy
```
wwood: this is a little vague - maybe give an example? Also, should mention that this actually gets modified (or change the code so it doesn't modify this variable).

geronimp: Is it preferable not to?

geronimp: In any case, ive added more information to the doco.

wwood: I don't get this 'v3' thing. Using `n` instead of `nm` is still valid v3 format.

This section describes the input hash, which makes 'stores' unclear. Stored before input or that's what this function does? Give an example of before and after I reckon. Why does 'place' need to be given as an empty array?

wwood: >Is it preferable not to?
In general it makes for less surprising code, yes

wwood: Also the `'0'` etc refers to sequence _sets_ not just sequences.

graftm/pplacer.py
```diff
@@ -147,7 +218,19 @@ def place(self, reverse_pipe, seqs_list, resolve_placements, files, args,
                 trusted_placements[base_file]={}
                 for read, entry in classifications[str(idx)].iteritems():
                     trusted_placements[base_file][read] = entry['placement']
+        # Split the original jplace file
+        # and write split jplaces to separate file directories
+        jplace_json = json.load(open(jplace))
```
wwood: rather
```
with open(jplace) as f: jplace_json = json.load(f)
```
so that the closing of the file handle is explicit

test/test_json_placement_split.py
```diff
@@ -0,0 +1,231 @@
...
+        
+        for alias_idx, alias_placement_entry in output_alias_hash.items():
+            expected_placement = expected_placements.pop()
+            placement_dict = alias_placement_entry['place']
+            observed_number_of_placements = len(placement_dict)
+            observed_placement = placement_dict[0]
+            self.assertEqual(expected_number_of_placements, 
+                             observed_number_of_placements)
+            self.assertEqual(expected_placement,
+                             observed_placement)
```
wwood: I still think this for loop complicates things, but OK

test/test_json_placement_split.py
```diff
@@ -0,0 +1,231 @@
...
+                             observed_number_of_placements)
+            self.assertEqual(expected_placement_results,
+                             observed_placement_results)
+            
+    def test_rereplicating_json_file(self):
+        input_alias_hash = {'0':{'place':[]}}
+        mock_cluster_hash = {
+                             '0':  {"test_read1": [Sequence("test_read1", "SEQUENCE"),
+                                                   Sequence("test_read2", "SEQUENCES")]}
+                             }
```
wwood: good good

wwood:
ok, good to go after the minor changes. Come see me about #171

PR #202, opened 2016-05-05
Bugfix: Do not attempt to re-create the alignment HMM when not all se…

geronimp:
…quences provided by the user align to the user-specified HMM

Very small change to address an issue tim was having when creating a GraftM package.


geronimp:
I'm going to merge this. The changes are only small anyway.

PR #203, opened 2016-05-07
reannotator: add failing unit test

wwood:
Another failing test (for me at least, and this is my code that is the culprit). I'm thinking that to fix this we move from skbio to dendropy, since rerooting we already have complex code for dealing with rerooting and it might just fall away with dendropy. Thoughts?


geronimp:
Agreed, I think it is best to move to dendropy

wwood:
ok to merge? I wrote it as skipped as a known failure.

geronimp:
Yes, I wasn't clear on whether you were ready, but yes all good.

Joel Boyd

PR #206, opened 2016-05-10
Dendropy reannotator

wwood:
Hey,

Now passing all the rerooting tests and all other nosetests too.

This should be backwards-compatible because `Reannotator._reroot_tree_by_old_root` has the same interface.

It actually makes more sense not to do things the 'pull request way' because the two commits do individual things - perhaps makes more sense to go through the commits separately and annotate anything you find in the commits themselves: e.g. write comments on these pages:
https://github.com/geronimp/graftM/commit/23c9afbd9d6348b1e912ce0ed37f1b9366afb7ef
https://github.com/geronimp/graftM/commit/636252d8b81ca41171f3bbc9406b60c703448566

because then you can see exactly what changed in the 2nd commit rather than seeing the move+change at the same time which is what you see if you just look at the '5 changed files'.

thanks for the review.


wwood:
Hey I fixed that bug I was talking about now, would you mind reviewing please? Thanks.

geronimp:
Hey, did you commit the bugfix?

wwood:
yes, with git commit --amend, git push -f, there's no extra commit

geronimp:
Okay, I see

Joel

wwood:
shall I merge?

PR #208, opened 2016-05-19
Create tweaks

wwood:
a few random things


graftm/run.py
```diff
@@ -608,14 +608,22 @@ def main(self):
...
                     exit(1)
-                
+
+                num_trees = 0
+                if self.args.rerooted_tree: num_trees += 1
+                if self.args.rerooted_annotated_tree: num_trees += 1
+                if self.args.tree: num_trees += 1
+                if num_trees > 1:
+                    logging.error("Only 1 input tree can be specified")
+                    exit(1)
```
geronimp: Could avoid creating another variable here by
```
if len(filter(None, [self.args.rerooted_tree, 
                          self.args.rerooted_annotated_tree, 
                          self.args.tree]) ):
           logging.error("Only 1 input tree can be specified")
           exit(1)
```
But I dont think it matters

graftm/sequence_io.py
```diff
@@ -8,7 +8,7 @@ def __init__(self, name, seq):
 
 class SequenceIO:
     # Stolen from https://github.com/lh3/readfq/blob/master/readfq.py
-    def _readfq(self, fp): # this is a generator function
+    def each(self, fp): # this is a generator function
```
geronimp: No problem, but just curious, why each?

graftm/create.py
```diff
@@ -830,9 +821,23 @@ def main(self, **kwargs):
...
+            for name, seq, _ in SequenceIO().each(open(sequences)):
+                count += 1
+                if count < 10:
+                    tf.write(">")
+                    tf.write(s.name)
+                    tf.write("\n")
+                    tf.write(s.seq)
+                    tf.write("\n")
+                else:
+                    break
```
geronimp: Sequence IO has a function for writing sequences: write_fasta_file
Do you mind using that to avoid redundant code?

geronimp:
Just a few comments. After looking at those if it passes nosetests im happy

graftm/sequence_io.py
wwood: `__iter__` is called `each` in Ruby, see.

graftm/run.py
wwood: ok

graftm/create.py
wwood: ok done

geronimp: Ok thanks, if you're happy you can merge

Joel

wwood:
these changes were merged after ensuring nosetests passes (some small changes were required).

PR #210, opened 2016-05-22
graftm_package: add is_protein_package method

wwood:
this is useful for singlem..


geronimp:
Would it not be better just to have this as a field in the CONTENTS.json file?

geronimp:
of the gpkg when its created that is

wwood:
yes, maybe, but I'm just too lazy to update all the gpkgs to another version. Shall we merge this and in v4 implement that as a field?

geronimp:
Sure

Joel

PR #214, opened 2016-05-30
rename bootstrap expand_search

geronimp:


wwood:
I think maybe you forgot to add a file to repo? btw you can use `git mv file1 file2` for situations like this.

geronimp:
Okay thanks for the tip, I've added the missing file

Joel

wwood:
OK if nosetests has nothing negative to say.

PR #215, opened 2016-05-31
Geronimp make merge reads default

geronimp:


bin/graftM
```diff
@@ -172,7 +172,7 @@ if __name__ == '__main__':
     pplacer_options = graft_parser.add_argument_group('pplacer assignment options')
     pplacer_options.add_argument('--placements_cutoff', metavar='confidence', help='Cutoff of placement confidence level (0.5 - 1), default=0.75', default=0.75)
     pplacer_options.add_argument('--resolve_placements', action="store_true", help='Force taxonomy down to the best placement (default=False)', default=False)
-    pplacer_options.add_argument('--merge_reads', action="store_true", help='Merge forward and reverse read alignments for placement (default=False)', default=False)
+    pplacer_options.add_argument('--dont_merge_reads', action="store_true", help='Merge forward and reverse read alignments for placement (default=False)', default=False)
```
wwood: `--no_merge_reads` I reckon, bit more standard even if it doesn't quite roll off the tongue.

graftm/run.py
```diff
@@ -241,7 +241,7 @@ def graft(self):
        
         # If merge reads is specified, check that there are reverse reads to merge with
         if self.args.merge_reads and not hasattr(self.args, 'reverse'):
-            logging.error("--merge requires --reverse to be specified")
+            logging.error("--merge_reads requires --reverse to be specified")
```
wwood: does this check need to happen? I'd just raise a Excception("Programming error") at this point, don't worry about a nice exit message because the user should not ever get here any more.

wwood:
ok other than that if nosetests check out

PR #222, opened 2016-08-30
geronimp-placement_cutoff_error

geronimp:


graftm/classify.py
```diff
@@ -42,8 +42,6 @@ def reduceTaxString(placement_hashes, threshold, resolve_placements):
                             cumil_confidence[item]+=normalised_confidences[idx]
                         else:
                             cumil_confidence[item]=normalised_confidences[idx]
-                    else:
-                        continue
```
wwood: ok, sure.

graftm/classify.py
```diff
@@ -68,10 +66,10 @@ def reduceTaxString(placement_hashes, threshold, resolve_placements):
                             break
                 else:    
                     best_place=max([(value, key) for key, value in cumil_confidence.items()])
-                    if best_place[0]>threshold:
```
wwood: Shouldn't the threshold be a float already? If not that sounds like a problem with the code calling this method, not this method, right? Looking through the code, I think that ultimately means adding `type=float, ` to this line: `pplacer_options.add_argument('--placements_cutoff', metavar='confidence', help='Cutoff of placement confidence level (0.5 - 1), default=0.75', default=0.75)` in `bin/graftM`

You could check by raising an exception if the cutoff is not already a float in the method you modified here.

Actually seems like there might be some other options that could do with similar treatment, like `--evalue`

wwood:
(and ok after that change).

Thanks.

geronimp:
Thanks Ben, I'll make those changes now

PR #223, opened 2016-09-06
Add graft --decoy_database.

wwood:


geronimp:
Just a thought, should sequences that were filtered by the decoy pipeline be retained somehow so that they can be added to the decoy database? It could help to develop a database of decoys as time goes on... Another option is to include decoys in the actual gpkgs, meaning you can "teach" the gpkg by adding them to the internal database. Would that be much effort to implement?

graftm/run.py
```diff
@@ -372,11 +378,26 @@ def graft(self):
                 if not result.hit_fasta():
                     logging.info('No reads found in %s' % base)
                     continue
-                elif self.args.search_only:
+
+                    
+
+                if self.args.search_only:
```
geronimp: Is this a fix? What happened here?

graftm/sequence_extractor.py
```diff
@@ -0,0 +1,41 @@
...
+                                   stdout=subprocess.PIPE)
+        process.communicate('\n'.join(reads_to_extract))
+
+    def extract_forward_and_reverse_complement(
+            self, forward_reads_to_extract, reverse_reads_to_extract, database_fasta_file,
+            output_file):
+        '''As per extract except also reverse complement the sequences.'''
+        self.extract(forward_reads_to_extract, database_fasta_file, output_file)
+        cmd_rev = "fxtract -XH -f /dev/stdin '%s' | " % database_fasta_file + \
+                  "seqmagick convert --reverse-complement --input-format fasta "\
```
geronimp: is seqmagick slower than just using fq2fa? 

probably too little a difference to care I guess..

geronimp:
Other than a few comments I've got no problems with this branch as it stands rn

Thanks for the good work

graftm/sequence_extractor.py
wwood: In this case `--reverse_complement` is also used so fq2fa won't work. I stole this code from singlem, this method actually isn't used in graftm.

graftm/run.py
wwood: since continue is there in the first if statement, the code is equivalent.

wwood:
Interesting idea about teaching, but not so easy to implement jsut now so I'm gonna leave it. Thanks for the review.

geronimp:
Thanks, didn't see your comments there - sorry for not getting to them

wwood:
All good, I wasn't expecting you to actually. I just wanted to do some further testing before pushing to master.

PR #224, opened 2016-09-15
Geronimp remove skbio from reroot

geronimp:


geronimp:
The commits aren't very modular, I apologize...

test/test_reroot.py
```diff
@@ -37,33 +36,38 @@
...
+        test_tree_1 =Tree.get(schema='newick', data=u'(A:0.1,B:0.2,(C:0.3,D:0.4):0.5);')
+        test_tree_2 =Tree.get(schema='newick', data=u'(A:0.5,B:0.2,(C:0.3,D:0.4):0.1);')
+        test_tree_3 =Tree.get(schema='newick', data=u'(A:0.2,B:0.5,(C:0.3,D:0.4):0.1);')
         
-        expected_test_tree_1 ="((A:0.1,B:0.2):0.25,(C:0.3,D:0.4):0.25)root:0;"
-        expected_test_tree_2 ="((B:0.2,(C:0.3,D:0.4):0.1):0.25,A:0.25)root:0;"
-        expected_test_tree_3 ="((A:0.2,(C:0.3,D:0.4):0.1):0.25,B:0.25)root:0;"
+        expected_test_tree_1 = str(Tree.get(schema='newick', data=u"((C:0.3,D:0.4):0.25,(A:0.1,B:0.2):0.25);"))
+        expected_test_tree_2 = str(Tree.get(schema='newick', data=u"(A:0.25,(B:0.2,(C:0.3,D:0.4):0.1):0.25);"))
+        expected_test_tree_3 = str(Tree.get(schema='newick', data=u"(B:0.25,(A:0.2,(C:0.3,D:0.4):0.1):0.25);"))
```
geronimp: Some of these expected trees needed re-arranging, but the overall topology is the same. The clades were just swapped around

geronimp:
Unfortunately there is still a TreeNode import in `test_tree_cleaner.py` because it imports `tree_cleaner.py` which I cant seem to see in the /graftm/ directory. do you knwo what is going on here?

test/test_reroot.py
wwood: sure, np

wwood:
I just removed test_tree_cleaner from the repo in the master branch. It was working for me because even though there was no `graftm/tree_cleaner.py` there was still a `graftm/tree_cleaner.pyc` file in my local directory. Gah, stupid python.

graftm/reannotator.py
```diff
@@ -1,28 +1,9 @@
...
-            data=str(skbio_tree),
-            schema='newick')
-
-    def dendropy_tree_to_skbio(self, dendropy_tree):
-        return TreeNode.read(
-            StringIO(unicode((str(dendropy_tree)+';').\
-                             replace(' ','_'))),
-            'newick')
+        r = Rerooter().reroot_by_tree(old_tree, new_tree)
+        return r
```
wwood: Can we just remove this file entirely? I'm not sure it makes sense to have a class with only a private method.

graftm/rerooter.py
```diff
@@ -61,50 +60,24 @@ def reroot(self, tree):
...
-                           if node != most_distant_node]
-            
-            dec_len = most_distant_node.length/2
-            
-            most_distant_node.length = dec_len
-            link_root = TreeNode(children=other_nodes, length = dec_len)
-            tree = TreeNode(children=[link_root, most_distant_node], 
-                            name = "root",
-                            length=0)
+            longest_edge_length = max(children.keys())
```
wwood: I suppose this achieves the aim of choosing an edge arbitrarily if there are 2 equally maximum edge lengths, so ok.

graftm/tree_decorator.py
```diff
@@ -130,21 +132,23 @@ def _rename(self, node, name):
...
         '''
-        if node.name:
+        if node.label:
             try: 
-                float(node.name)
-                node.name = "%s:%s" % (node.name,
+                float(node.label)
+                new_label = "%s:%s" % (node.label,
                                        name)
             except:
```
wwood: can we add a specific exception here?

geronimp:
Ah, that must be why it was working for me too. So why was it removed?

wwood:
It became dendropy_tree_cleaner because that it was helpful to have both around.

geronimp:
ok cool in that case this branch should be good to go if you're happy with my changes.

wwood:
You missed one in decorate.py I think

geronimp:
Oops. doesnt appear to be used though. I'll fix that now

geronimp:
try now?

wwood:
I didn't get any skbio errors in guix after applying this tiny commit https://github.com/wwood/graftM/commit/d0a8f66ef93f8394a7b52f7c58732cd68d35616a

But there are some other strange errors. I believe these are more to do with guix than graftm though, there's something odd going on with tempdirs.

It also fails because of diamond db incompatibility issues. Perhaps for 0.10.0 we should reformat the diamond dbs to the new version and update the readme. I've set it to use diamond 0.7.9 so there's no issues.

But, caveats aside, we have a working graftm guixified

:)

geronimp:
hurrah!

Joel

wwood:
pushed all these commits to master just now

PR #225, opened 2016-09-20
Final skbio removal

wwood:
This is to merge into your remove_skbio_from_reroot.

I've also rebased so there's some extraneous code in there too sorry. But the last 5 commits are the important bit. After this I think we are good to merge both our PRs into master.


wwood:
Hey, I keep hacking on this branch, shall we merge the whole thing? There are 2 new commits today due to Caitlin's diamond threads error.

PR #228, opened 2016-10-12
Updated License information and added SILVA license

geronimp:


wwood:
I think that looks fine, except for the (currently meaningless) detail that it is under GPL 3+, not GPL 3.

PR #231, opened 2017-01-26
test: Update diamond DBs to a post 0.7.9 format.

wwood:
I can't remember where we are up to with diamond, but I think we should update the DBs to the newer version before pushing out the next version. The distributed gpkgs will have to be updated too, I guess.


wwood:
Ah sorry this only passes the graft tests, some others now fail. Don't worry for now.

wwood:
OK now all the nosetests pass, ready for your review.

PR #239, opened 2017-07-12
Bug fix for issue 198

geronimp:


graftm/housekeeping.py
```diff
@@ -199,5 +199,10 @@ def set_attributes(self, args):
                 raise Exception("Specified HMM search_hmm_files when not using the hmmsearch pipeline. Using: %s" % (args.search_method))
 
         else:
-            raise Exception('No refpkg or HMM specified: Do not know what to search with.')
+            if args.search_only:
+                if args.search_diamond_file:
+                    args.search_method = 'diamond'
```
wwood: Use `Run.DIAMOND_SEARCH_METHOD`. Probly change it in the argparse too.

graftm/housekeeping.py
```diff
@@ -199,5 +199,10 @@ def set_attributes(self, args):
...
+                    args.search_hmm_files = [None]
```
wwood: Why not `[]`

graftm/housekeeping.py
```diff
@@ -199,5 +199,10 @@ def set_attributes(self, args):
...
+            else:
+                raise Exception('No refpkg or HMM specified: Do not know what to search with.')
```
wwood: `No refpkg` ? What has that to do with searching? Shouldn't mention a diamond DB here?

PR #243, opened 2018-01-19
GraftM can now take >7 taxonomy ranks when creating a graftM package

geronimp:


graftm/__init__.py
```diff
@@ -1,8 +1,4 @@
-""" graftm/__init__.py """
-
-# __all__ = []
-
-#from .graftm import *
+""" enrichm/__init__.py """
```
wwood: not sure this is right

geronimp: not sure how this slipped through. Thanks

PR #245, opened 2018-02-09
Remove seqmagick

wwood:
Since seqmagick has become python-3, thought we should remove it. Easy enough since it's a thin wrapper around Bio::SeqIO. Also should make graftm slightly faster as fewer subprocess python instances need to be started.

Apologies for the whitespace noise. I'm reasonably confident here as all the changes are quite local within functions, and the tests all pass.

### DIFF
--- a/graftm/create.py
+++ b/graftm/create.py
@@ -5,6 +5,7 @@ import shutil
 import tempfile
 import logging
 import subprocess32
+import subprocess
 import extern
 import tempdir
 import json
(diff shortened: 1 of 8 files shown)


geronimp:
Looks good, although the stdout is lost in of some parts (used to be caught by extern and reported via logging.debug).

I think it should be fine though. Feel free to merge if you're happy, thanks Ben

PR #247, opened 2018-03-12
graft: Use specified --input_sequence_type.

wwood:
I came across a bug - the input sequence type was being ignored as a parameter. Most of the time the guessing works fine so it wasn't an issue. Fixing that also means things are slightly faster when it is specified (1% on my laptop), which is a little nice.


wwood:
merged

PR #257, opened 2018-12-21
Fixes dec2018

wwood:
Hey @geronimp here's a few fixes. The main thing is ba6f11f which handles #254. The difficulty was that when matching reads are very rare in a sample, it is possible to have a forward aln file generated, but not the reverse (or vice versa, and also some may be filtered out rather than not picked up at all in the first place). To fix this I decided on the somewhat lazy approach of making a .aln.fa file for each read set, even if it was empty. That way the array of aln filenames has an even length, and they all match up, etc. and it doesn't require a large rewrite of run.py. The change to the user then is that there may be empty .aln.fa files in the final output folder. Does that make sense to you? Maybe, we could specifically remove these files at the end.

OK with you?


geronimp:
Hey @wwood 

This looks totally fine to me, and leaving the file is probably fine. Feel free to merge if you're happy

Merry Christmas

wwood:
thanks, merry xmas to you too.

PR #272, opened 2021-09-09
add create no-tree flags

AroneyS:
Forgot to add the actual flag for create --no-tree.
Is there anywhere else I need to add it?


wwood:
I don't think so, graftM's only semi-stable interface is the cmdline.

PR #280, opened 2022-05-30
Remove backports.tempfile dependency

AroneyS:


wwood:
This can't be merged yet because it would break SingleM dev I think. SingleM dev branch currently relies on Python 3.6, because of a dependency that needs some admin coding to get rid of.

Is not merging this causing problems somehow?

AroneyS:
I don't have any reason to merge it yet, just forgot about it.

wwood:
Actually seems graftm's bioconda recipe has 3.7+ already, so just merging so don't have to remember later.

